PR #4, opened 2020-04-24
Updated examples

takidau:
- Fix role name in playbook
- Add token var, since it's required
- Add "become: yes" since sudo is needed
- Add an example Inventory file, since "private_ip" and "id"
  vars are required.


senior7515:
LGTM.

PR #6, opened 2020-10-26
Automatically detect when raid is needed

rkruze:
In the past, you had to specify if the instance was a raid or not raid. It now looks to see how many drives are listed which have no partitions. If this is larger than 1, then we raid is created. Otherwise, the disk is just formatted and mounted.


0xdiba (approved):
:rocket: 
https://github.com/vectorizedio/deployment-automation/blob/3d1497d1a9e1aa0d6695a6c5abf9a794b9c1f514/ansible/playbooks/provision-node.yml#L3-L7
Let's change this to just `- include: prepare-data-dir.yml` since the `with_raid` var is now useless and merge!

rkruze:
Now raid doesn't have to be specified when deploying a cluster. It will auto-detect.

PR #9, opened 2020-11-06
Auto generate the host.ini needed for ansible from terraform

rkruze:


ansible/playbooks/provision-node.yml
```diff
@@ -5,7 +5,5 @@
 - include: install-redpanda.yml
 
 - include: start-redpanda.yml
-  when: start | default(False)
```
0xdiba, line 8: I'd say let's remove this commit from this PR and merge them separately.

0xdiba (commented):
Awesome! Let's remove the non-relevant commit and :rocket:

PR #13, opened 2020-11-11
Update GCP readme

0x5d:


0x5d:
I updated the readme, which can be merged. However, the Ansible step hangs when deploying the monitor, and I haven't been able to find out why.

PR #15, opened 2020-12-22
Set default for advertise_public_ips and use packages.vectorized.io

rkruze:
There was no default for advertise_public_ips, so it would fail unless
the user specified this parameter. Also updated to use the new package
repo and drop the requirement for a token.


0xdiba (approved):
Nice catch! thanks for taking care of this

PR #20, opened 2021-01-21
Added the port for node exporter.

rkruze:


aws/cluster.tf
```diff
@@ -106,6 +106,14 @@ resource "aws_security_group" "node_sec_group" {
     cidr_blocks = ["0.0.0.0/0"]
   }
 
+  # node exporter
+  ingress {
+    from_port   = 9100
```
0xdiba, line 111: Are we sure we want to expose node exporter to the world? the info can probably be scraped by the prometheus instance.

Not that it makes any difference security wise ( the cluster is unprotected anyway ), just curious on the usecase.

rkruze: Fair point, it doesn't work without this rule, but perhaps we should update all things that need it to use the security group name vs 0.0.0.0/0.

rkruze: Ok, changed it, we now open these ports to the security group and not to the full internet.

0xdiba (approved):
Looks good! :rocket:

PR #30, opened 2021-07-09
Ansible to enable TLS on Redpanda

patrickangeles:


ansible/playbooks/tls/install-certs.yml
```diff
@@ -0,0 +1,58 @@
...
+      dest: /etc/redpanda/certs/node.crt
+      owner: redpanda
+      group: redpanda
+
+  - name: Configure via RPK
+    shell:
+      cmd: |
+        rpk redpanda config set redpanda "
+
+          kafka_api:
```
senior7515, line 23: 🔥 

this is awesome. @0x5d  - check it out :D

ansible/playbooks/tls/templates/ca.conf.tpl
```diff
@@ -0,0 +1,49 @@
+# OpenSSL CA configuration file
```
senior7515, line 1: this is great.

senior7515 (commented):
Looks great! i'll wait for @0xdiba and @ivotron to chime in.

0xdiba (approved):
Very Nice! LGTM
As a next step we can add it as a var ( eg `with_tls`) and do the process along with everything else when deploying the nodes.

Thanks for tackling this!

ivotron (approved):
lgtm!

echoing @0xdiba comment, the only issue is that running the `install-certs` playbook overwrites the config, which might have some unintended consequences.

ansible/playbooks/tls/install-certs.yml
0x5d: Great! 😊

ansible/playbooks/tls/install-certs.yml
```diff
@@ -0,0 +1,58 @@
...
+            require_client_auth: false
+            cert_file: /etc/redpanda/certs/node.crt
+            key_file: /etc/redpanda/certs/node.key
+            truststore_file: /etc/redpanda/certs/truststore.pem
+
+        " --format yaml
+        rpk redpanda config set rpk "
+            tls:
+              truststore_file: /etc/redpanda/certs/truststore.pem
+        " --format yaml
```
0x5d, line 52: @patrickangeles this changed last week 😁 While this is still supported, the recommended field is now `rpk.kafka_api.tls`:
```suggestion
        rpk redpanda config set 'rpk.kafka_api' "
            tls:
              truststore_file: /etc/redpanda/certs/truststore.pem
        " --format yaml
```

This is so the Admin API TLS can be configured separately (`rpk.admin_api.tls`)

0x5d (changes requested):

ansible/playbooks/tls/issue-certs.yml
```diff
@@ -0,0 +1,17 @@
+- hosts: redpanda
```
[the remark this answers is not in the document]
patrickangeles: It's split up into generate-csr, issue-certs, and install-certs because the middle step (issue-certs) could be done externally, using a corporate CA or whatever...

PR #35, opened 2021-09-10
aws: provisioning for client VMs and README tweaks

ajfabbri:
- aws: readme tweaks.
- asw: readme tweaks
- aws: add AMI id for ubuntu hirsute build.
- aws: add support for provisioning blank client VMs

### DIFF
--- a/aws/vars.tf
+++ b/aws/vars.tf
@@ -21,7 +21,7 @@ variable "instance_type" {
 
 variable "client_instance_type" {
   description = "Default client instance type to create"
-  default     = "m5n.8xlarge"
+  default     = "m5n.2xlarge"
 }
 
 variable "prometheus_instance_type" {
--- a/aws/vars.tf
+++ b/aws/vars.tf
@@ -49,6 +49,7 @@ variable "distro_ami" {
     # https://cloud-images.ubuntu.com/locator/ec2/
     "ubuntu-focal"  = "ami-02c45ea799467b51b"
     "ubuntu-bionic" = "ami-0c1ab2d66f996cd4b"
+    "ubuntu-hirsute" = "ami-035649ffeb04ce758"  # non-LTS for development
 
     # https://us-west-2.console.aws.amazon.com/ec2/v2/home?region=us-west-2#LaunchInstanceWizard:
     "rhel-8"         = "ami-087c2c50437d0b80d"
@@ -66,6 +67,7 @@ variable "distro_ssh_user" {
     "fedora-32"      = "fedora"
     "ubuntu-bionic"  = "ubuntu"
     "ubuntu-focal"   = "ubuntu"
+    "ubuntu-hirsute" = "ubuntu"
     "rhel-8"         = "ec2-user"
     "amazon-linux-2" = "ec2-user"
   }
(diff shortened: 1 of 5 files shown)


dotnwat (approved):
looks good to me. let's let @rkruze or @ivotron have a peek since I think they are the primary authors?

aws/vars.tf
```diff
@@ -19,6 +19,11 @@ variable "instance_type" {
   default     = "i3.2xlarge"
 }
 
+variable "client_instance_type" {
+  description = "Default client instance type to create"
+  default     = "m5n.8xlarge"
```
rkruze, line 24: This is a very beefy default client node, I would recommend something like an m5n.2xlarge.

aws/vars.tf
```diff
@@ -30,6 +35,17 @@ variable "enable_monitoring" {
   default     = true
 }
 
+variable "clients" {
+  description = "Number of kafka client hosts to set up, if any."
+  type        = number
+  default     = 0
+}
+
+variable "client_distro" {
```
rkruze, line 44: Is there a reason to have a different distro for the client vs the rest of the cluster?

rkruze (changes requested):
It would also be good if rpk was installed on the client machine so that it could be useful for creating/deleting topics.

aws/vars.tf
ajfabbri: Ok, thanks. I picked the smallest one that had guaranteed 25 Gbps network throughput, but that is probably overkill for most basic testing. I'll add another commit to change the default here.

aws/vars.tf
ajfabbri: In the case that some tools/tests are easier to build on a different (newer) distro. Default is the same distro though.

ajfabbri:
> It would also be good if rpk was installed on the client machine so that it could be useful for creating/deleting topics.

I don't add any setup of the client machines here. As mentioned in the README and commit message, these are just blank VMs for now. We probably want to add some basic ansible roles for clients here. For now I'm using other ansible scripts once I finish my VM provisioning.

dotnwat:
> It would also be good if rpk was installed on the client machine so that it could be useful for creating/deleting topics.

@ajfabbri i think that in the ansible scripts in this project we can install the redpanda package on the client nodes

the redpanda-ansible client-testing role in the other repository shouldn't install rpk, rather we can create a resuable redpanda role at some point.

aws/vars.tf
dotnwat: would it make sense to use something like c5.4xlarge or something? the m5n nodes seem to have a lot of local storage. it's not clear we need that in the default case

dotnwat:
i tried this an it works. what is left before this can be merged?

aws/vars.tf
ajfabbri: I'd say `c5n.xlarge` when 4 vCPUs is enough, and `m5n.2xlarge` when you want 8 cores + more memory:
image

ajfabbri:
I have another patch and a wiki writeup waiting for this. I'm assuming y'all don't want me to take the time *now* to factor out a separate RPK installer / role and use it for the blank client machines. As is rpk is available on the RP nodes. These ansible scripts install all of RP and get it that way.

rkruze (approved):

PR #40, opened 2021-11-08
Azure Terraform Configuration

jrkinley:


patrickangeles (approved):
thanks for this!

azure/cluster.tf
```diff
@@ -0,0 +1,140 @@
...
+# Redpanda broker VMs
+#
+
+resource "azurerm_linux_virtual_machine" "redpanda" {
+  name                 = "redpanda_broker${count.index}"
+  computer_name        = "redpanda${count.index}"
+  count                = var.vm_instances
+  resource_group_name  = azurerm_resource_group.redpanda.name
+  location             = azurerm_resource_group.redpanda.location
+  # availability_set_id  = azurerm_availability_set.redpanda.id
```
rkruze, line 11: Is there a reason this is commented out?

azure/cluster.tf
```diff
@@ -0,0 +1,140 @@
...
+  }
+}
+
+resource "azurerm_managed_disk" "redpanda" {
+  name                 = "managed_disk${count.index}"
+  count                = var.vm_add_data_disk ? var.vm_instances : 0
+  resource_group_name  = azurerm_resource_group.redpanda.name
+  location             = azurerm_resource_group.redpanda.location
+  storage_account_type = "Premium_LRS"
+  # storage_account_type = "UltraSSD_LRS"
```
rkruze, line 49: Should this be removed?

rkruze (approved):
Just a few comments on my end! This is great!

azure/cluster.tf
jrkinley: Due to general lack of availability it’s difficult to acquire Lsv2 instances in the same availability set, which causes Terraform to constantly receive AllocationFailed error messages. Commenting this out widens the hardware Azure can select from (meaning that Lsv2 instances can be deployed on the same server). I'll put this back in for now as ideally we want the brokers to be in an availability set to benefits from the 99.95% SLA.

PR #43, opened 2021-11-12
Prevent deployment ID from changing every plan

### LINKED ISSUE #42: Terraform AWS deploy always regerates all resources
Every time you run a terraform command, almost every resource (all the hosts, in particular) are torn down and replaced, so al the IP addresses change, any existing state is lost, etc. 

The plan shows that the cause is the use of `timestamp()` as part of the `deployment_id`. Unlike the random UUID which will change only when the deployment is initially created, the timestamp is different on _every_ call, meaning that the deployment ID changes every time and we can't apply changes to an existing deployment: it effectively creates a new deployment every time. 

As a minimal example, if you run two consecutive `terraform apply` with the AWS modules, you would expect the second does nothing: after all, nothing has changed and the first apply put everything into the desired state. However, currently all hosts will be torn down and new ones created. The only thing that changed was the ssh key name, which is based on the deployment ID.

travisdowns:
The use of timestamp() as part of the deployment ID means that,
in effect, the deployment ID changes every time a plan is created.
Instead, we use time_static, which is the recommended way to
introduce timestamps into a resouce: the timestamp is genenated only
when the resource is created in the first place.

Fixes #42.

### DIFF
--- a/aws/cluster.tf
+++ b/aws/cluster.tf
@@ -1,9 +1,11 @@
 resource "random_uuid" "cluster" {}
 
+resource "time_static" "timestamp" {}
+
 locals {
-  timestamp = timestamp()
   uuid = random_uuid.cluster.result
-  deployment_id = "redpanda-${random_uuid.cluster.result}-${local.timestamp}"
+  timestamp = time_static.timestamp.rfc3339
+  deployment_id = "redpanda-${local.uuid}-${local.timestamp}"
 }
 
 resource "aws_instance" "redpanda" {


aws/cluster.tf
```diff
@@ -1,9 +1,11 @@
 resource "random_uuid" "cluster" {}
 
+resource "time_static" "timestamp" {}
```
travisdowns, line 3: This time_static resources will capture the timestamp when it is first created, but won't change after that until the resource is destroyed, which I believe is the desired behavior (and the recommended way to handle timestamps in terraform).

0xdiba (approved):

ivotron (approved):

PR #46, opened 2021-11-12
Create a tag with IAM user on resources.

### LINKED ISSUE #45: Hard to tell who created instances
It's hard to tell who created a given node, i.e., to find abandoned machines or so users can check what they have running.

I propose we add a tag with the IAM username to make it clearer.

travisdowns:
A tag iam_username is created on resources such as instances with
name of the IAM user who ran terraform to create the resource.

Fixes #45.

### DIFF
--- a/aws/cluster.tf
+++ b/aws/cluster.tf
@@ -6,6 +6,12 @@ locals {
   uuid = random_uuid.cluster.result
   timestamp = time_static.timestamp.rfc3339
   deployment_id = "redpanda-${local.uuid}-${local.timestamp}"
+
+  # tags shared by all instances
+  instance_tags = {
+    owner        : local.deployment_id
+    iam_username : trimprefix(data.aws_arn.caller_arn.resource, "user/")
+  }
 }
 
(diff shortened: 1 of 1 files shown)


aws/cluster.tf
```diff
@@ -6,6 +6,12 @@ locals {
   uuid = random_uuid.cluster.result
   timestamp = time_static.timestamp.rfc3339
   deployment_id = "redpanda-${local.uuid}-${local.timestamp}"
+
```
travisdowns, line 9: Since we use the same tags on every resources, I factored them out into a local.

aws/cluster.tf
```diff
@@ -151,3 +149,12 @@ resource "local_file" "hosts_ini" {
   )
   filename = "${path.module}/../hosts.ini"
 }
+
```
travisdowns, line 152: This is the best way I could find to get the IAM username.

ajfabbri (approved):
LGTM

ivotron (approved):

Issue #49, opened 2021-11-13
Steps with multiple-line shell scripts only catch failures on the last command


travisdowns:
Consider the configure redpanda step:

~~~yaml
- name: configure redpanda
    notify:
      - restart redpanda-tuner
      - restart redpanda
    vars:
      use_public_ips: "{{ advertise_public_ips | default(false, true) | bool }}"
    shell: |
      rpk config set cluster_id 'redpanda'
      rpk config set organization 'redpanda-test'
      rpk config set redpanda.advertised_kafka_api '{
      {% if use_public_ips %}
        address: {{ inventory_hostname }},
      {% else %}
        address: {{ hostvars[inventory_hostname].private_ip }},
      {% endif %}
        port: 9092
      }' --format yaml
      rpk config set redpanda.advertised_rpc_api '{
        address: {{ hostvars[inventory_hostname].private_ip }},
        port: 33145
      }' --format yaml
      sudo rpk mode production

      {% if hostvars[groups['redpanda'][0]].id == hostvars[inventory_hostname].id %}
      sudo rpk config bootstrap \
        --id {{ hostvars[inventory_hostname].id }} \
        --self {{ hostvars[inventory_hostname].private_ip }}

      {% else %}

      sudo rpk config bootstrap \
        --id {{ hostvars[inventory_hostname].id }} \
        --self {{ hostvars[inventory_hostname].private_ip }} \
        --ips {{ groups["redpanda"] | map('extract', hostvars) | map(attribute='private_ip') | join(',') }}
      {% endif %}
~~~

If any `shell` command fails other than the last one, this step will succeed when it should fail, since these steps are effectively run as a shell script whose return value is that of the last command to run.

The easiest fix is probably just a `set -e` at the top of the script.

PR #52, opened 2021-11-13
Use set -e in multi-line shell commands

travisdowns:
Otherwise, failures in anything other than the last line wil
be silently ignored.

Issue #49.


ansible/playbooks/start-redpanda.yml
```diff
@@ -13,6 +13,7 @@
     vars:
       use_public_ips: "{{ advertise_public_ips | default(false, true) }}"
     shell: |
+      set -e
```
travisdowns, line 16: Tested locally and confirmed this will catch failures.

ivotron (approved):

PR #62, opened 2022-06-15
Remove java/kafka, update grafana version and rpk command

larsenpanda:
Remove Kafka binaries, remove jdk install, change grafana to latest and update --prometheus-url to --metrics-endpoint

### DIFF
--- a/ansible/playbooks/deploy-prometheus-grafana.yml
+++ b/ansible/playbooks/deploy-prometheus-grafana.yml
@@ -29,7 +29,7 @@
   roles:
   - cloudalchemy.grafana
   vars:
-    grafana_version: 7.5.10
+    grafana_version: latest
     grafana_security:
       admin_user: admin
       admin_password: "{{ grafana_admin_pass | default('enter_your_secure_password', true) }}"
--- a/ansible/playbooks/deploy-prometheus-grafana.yml
+++ b/ansible/playbooks/deploy-prometheus-grafana.yml
@@ -2,7 +2,7 @@
   tasks:
   - name: generate the redpanda grafana dashboard
     shell: |
-      rpk generate grafana-dashboard --datasource prometheus --prometheus-url 'http://{{hostvars[inventory_hostname].private_ip}}:9644/metrics' > '/tmp/redpanda-grafana.json'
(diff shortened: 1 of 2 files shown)


ivotron (commented):
would it be possible to add an explanation for the removal of Kafka binaries?

larsenpanda:
Sure, the reason those were there, according to Roko, are tests that are no longer necessary, and there is 2 problems with having them - a) it requires installing java which we don't want to install on any redpanda machines unless absolutely necessary, and b) they're getting flagged by vulnerability checkers for the Log4jShell CVE, and cflt won't be changing the log4j version to get past log4jshell issue for a while, they already told all of their customers that. During our POC the prospect immediately found the issue and burned the cluster to the ground. It's more trouble than it's worth in my (and Roko's) opinion.

ansible/playbooks/deploy-prometheus-grafana.yml
```diff
@@ -29,7 +29,7 @@
   roles:
   - cloudalchemy.grafana
   vars:
-    grafana_version: 7.5.10
+    grafana_version: latest
```
rkruze, line 32: Just to check, have you verified that our grafana dashboard works with the latest version?

larsenpanda: I pulled up grafana and imported the dashboard and saw that everything looked fine. BUT i had to go through the process of setting up grafana and prometheus manually ahead of time and had nothing going on the cluster. The charts were empty which was what i expect, and figured importing the charts json that was generated would be what caused a problem, so without doing additional testing of creating topics, producing/consuming, and it not giving an error after hooking up prometheus I deemed the addl steps unnecessary. LMK if you want me to do addl testing.

larsenpanda: ok, i'll look at what their latest version number is and try it out. I'm also in the process of having prometheus/grafana installed along with it, and will put some load on it to make sure the charts draw.

larsenpanda: TIL that i should have expected grafana and prometheus to get installed and dashboard automagically imported OOB, but didn't b/c I was trying to use the same host for both monitoring and redpanda. I am re-running tests in this fork, stand by..

larsenpanda: changed to and tested with pinning to Grafana *9.0.0* (latest) and it deployed fine, imported fine, all charts rendered actively.

PR #63, opened 2022-06-28
Disable qdc by default, make it configurable

travisdowns:
Curently QDC (queue depth control) is enabled by default, but as
this can be detrimental to performance this we now disable it by
default. However, it can be enabled again with the existing
values by setting enable_qdc true when running the
playbook.


ivotron (approved):

travisdowns:
This is closed because we did https://github.com/redpanda-data/deployment-automation/pull/64 instead.

PR #74, opened 2022-09-13
Support rack aware deployment across Azure and GCP

tmgstevens:
When ha variable is set (defaults to false) will create a partition placement group and ensure that nodes are spread between the groups


vuldin (commented):
My only reason for not yet approving is that I haven't tested these changes out across multiple cloud providers (I need to get access to Azure, for instance).

azure/vars.tf
```diff
@@ -19,6 +19,12 @@ variable "vm_instances" {
   default     = 3
 }
 
+variable "use_scale_set" {
```
r-vasquez, line 22: Can we update the READMEs with the new vars?

ansible/playbooks/start-redpanda.yml
```diff
@@ -45,6 +46,16 @@
...
+      {% if rack is defined and rack != -1 %}
+      sudo rpk redpanda config set redpanda.rack {{ rack }}
+      {% endif %}
+
+  - name: enable rack awareness
+    shell: |
+      set -e
+      rpk cluster config set enable_rack_awareness true --api-urls {{ hostvars[inventory_hostname].private_ip }}:9644
+
+    when: rack is defined and rack != -1
```
r-vasquez, line 58: [Suggestion] what do you think of leaving these 2 together:

```yaml
  - name: enable rack awareness
    shell: |
      set -e
      sudo rpk redpanda config set redpanda.rack {{ rack }}
      rpk cluster config set enable_rack_awareness true --api-urls {{ hostvars[inventory_hostname].private_ip }}:9644
    when: rack is defined and rack != -1
```

ansible/playbooks/start-redpanda.yml
```diff
@@ -45,6 +46,16 @@
...
       {% endif %}
 
+      {% if rack is defined and rack != -1 %}
+      sudo rpk redpanda config set redpanda.rack {{ rack }}
+      {% endif %}
+
+  - name: enable rack awareness
+    shell: |
+      set -e
+      rpk cluster config set enable_rack_awareness true --api-urls {{ hostvars[inventory_hostname].private_ip }}:9644
```
r-vasquez, line 56: The `--api-urls` is a comma-separated list of admin API addresses (IP:port) but this will give you a single address, can we change this so we map each IP to a list (see line 46)?

[question], Should we add a conditional so we use the `inventory_hostname` if "`use_public_ips`" is set?

vuldin (changes requested):
Rogger and I met to discuss this PR yesterday and had some questions/concerns:
- the use of the proximity placement group seems odd, from reading the docs it seems this is meant to do the opposite of splitting nodes up across racks and increasing availability
- there isn't very much consistency in the variable names across the aws/azure/gcp code sets
- we ran into issues with it working as expected when deployed to Azure (I can't recall the exact issue but will post more details)

At the same time the AWS side worked well with a minor change. Since a customer is wanting to have this soon, we thought breaking out the AWS part into it's own PR may be the best approach. More details in this PR: https://github.com/redpanda-data/deployment-automation/pull/85

ansible/playbooks/start-redpanda.yml
tmgstevens: @vuldin has addressed this

ansible/playbooks/start-redpanda.yml
tmgstevens: @vuldin also addressed this

tmgstevens:
@vuldin @r-vasquez regarding your comments above:

> the use of the proximity placement group seems odd, from reading the docs it seems this is meant to do the opposite of splitting nodes up across racks and increasing availability

The proximity placement group stuff was already in there - I think ideally we want nodes to be as close together as possible (for latency) but respecting the scale-set failure domain constraints. The only risk here is that we get availability problems by pairing them both together. If we see that we'll have to consider making it an either/or.

> there isn't very much consistency in the variable names across the aws/azure/gcp code sets

I agree, at this point the three terraform implementations are pretty disparate. I'm not really fancying a rework, but we should make sure that any new parameter names are consistent. I'll check that.

> we ran into issues with it working as expected when deployed to Azure (I can't recall the exact issue but will post more details)

Grateful for info. It's worked fine for me so far.

azure/cluster.tf
```diff
@@ -8,8 +8,11 @@ resource "azurerm_linux_virtual_machine" "redpanda" {
   count                        = var.vm_instances
   resource_group_name          = azurerm_resource_group.redpanda.name
   location                     = azurerm_resource_group.redpanda.location
-  availability_set_id          = azurerm_availability_set.redpanda.id
+  availability_set_id          = var.ha ? null : azurerm_availability_set.redpanda.0.id
```
vuldin, line 11: So if `ha` is `true`, then we don't create an availability set? That seems opposite of what we want right?

azure/network.tf
```diff
@@ -25,10 +25,24 @@ resource "azurerm_proximity_placement_group" "redpanda" {
 }
 
 resource "azurerm_availability_set" "redpanda" {
-  name                         = "redpanda_availability_set"
+  name                         = "redpanda_availability_set_${local.deployment_id}"
   resource_group_name          = azurerm_resource_group.redpanda.name
   location                     = azurerm_resource_group.redpanda.location
   proximity_placement_group_id = azurerm_proximity_placement_group.redpanda.id
+  count                        = var.ha ? 0 : 1
```
vuldin, line 32: Same questions as above: this value seems opposite to what I would expect we would want (to create an availability set when `ha` is `true`).

vuldin (commented):
I think I don't have a clear understanding of how availability sets, placement groups, and scale sets work with each other to enable high availability. From reading the docs I would expect that we want to create an availability set whenever `ha` is `true`.

azure/vars.tf
tmgstevens: Updated.

ansible/playbooks/start-redpanda.yml
tmgstevens: Now resolved on main branch

azure/cluster.tf
tmgstevens: So the _old_ behaviour for this provider is that we would automatically deploy machines in an availability set, presumably for resilience. However availability sets don't give the ability to introspect which machines are placed into which fault domain, so there is limited use in terms of rack awareness in Redpanda. We _could_ remove availability sets altogether, other than it would be a change in behaviour from what was there before.

When `ha=true` we now deploy into a flexible scale set (with three fault domains) which then gives us information about which fault domain each VM is, that we can use for rack awareness. 

I figured there was no harm in leaving the existing behaviour in there, but happy to take advice.

tmgstevens:
Updated addressing comments. @vuldin be good to get a re-review please.

azure/cluster.tf
vuldin: Thanks, this helps explain the use of availability sets in this project. Scale set (rather than availability set) is the vehicle for enabling rack awareness, and I was mistakenly trying to understand how the availability set could provide that especially given changes in this PR.

vuldin (approved):
Thanks Tristans, looks good. I see `start-redpanda.yml` is now included as an empty file in this PR, I think that file could be deleted entirely.

tmgstevens:
Thanks @vuldin

Issue #82, opened 2022-10-18
Enable rack awareness


r-vasquez:
The user should be able to opt-in for rack awareness with an ansible variable:  `-e enable-rack-awareness=true`

See: https://github.com/redpanda-data/deployment-automation/issues/65#issuecomment-1236692074

vuldin:
Related issue: https://github.com/redpanda-data/redpanda/issues/6827

WesWWagner:
Is this not complete via the support of the rack ids in hosts.ini ?

WesWWagner:
This feature already exists in the current main line code base.

set rack=# in the hosts ini or extra vars , and rack awareness will be enabled and assigned to the correct rack ID.

WesWWagner:
Already in /main as a feature

PR #85, opened 2022-10-21
Enable high availability and rack awareness on AWS

### LINKED ISSUE #80: jmespath install is needed prior to running json_query_filter
When running the provision node ansible playbook, the redpanda nodes succeed but the prometheus node fails with the following message:

```
TASK [cloudalchemy.prometheus : Get all file_sd files from scrape_configs] *************************************************************************************************************************************************************
fatal: [35.84.208.32]: FAILED! => {"msg": "You need to install \"jmespath\" prior to running json_query filter"}
```

vuldin:
Tristan has an initial PR for implementing HA and rack awareness across AWS, GCP, and Azure. But during testing we ran into some issues in Azure that may require changes. At the same time a customer requires rack awareness in AWS.

This PR focuses on implementing this feature only in the AWS implementation. This will give us time to work through the changes needed for other cloud providers without impacting AWS customers. #82 is not closed with this PR as that ticket includes all cloud providers.

I've also included a small fix for prometheus so it properly starts (a dependency was missing from the Ubuntu distro).

In order to test, change this variable to true: https://github.com/redpanda-data/deployment-automation/blob/1a8f8f3fd799808b3acc3a06989d64abc91b6d9f/aws/vars.tf#L12-L16

Then run the following commands from the root directory:
```
cd aws
terraform init
terraform apply
cd ..
ansible-playbook --private-key `cat ~/.ssh/id_rsa.pub | awk '{print $2}'` -i hosts.ini -v ansible/playbooks/provision-node.yml -e advertise_public_ips=true -e grafana_admin_pass=password
```

fix #80

### DIFF
--- a/ansible/playbooks/start-redpanda.yml
+++ b/ansible/playbooks/start-redpanda.yml
@@ -7,16 +7,13 @@
       state: directory
 
   - name: configure redpanda
-    notify:
-      - restart redpanda-tuner
-      - restart redpanda
     vars:
       use_public_ips: "{{ advertise_public_ips | d() | bool }}"
     shell: |
       set -e
-      rpk config set cluster_id 'redpanda'
-      rpk config set organization 'redpanda-test'
-      rpk config set redpanda.advertised_kafka_api '{
+      sudo -u redpanda rpk redpanda config set cluster_id 'redpanda'
+      sudo -u redpanda rpk redpanda config set organization 'redpanda-test'
+      sudo -u redpanda rpk redpanda config set redpanda.advertised_kafka_api '{
       {% if use_public_ips %}
         address: {{ inventory_hostname }},
       {% else %}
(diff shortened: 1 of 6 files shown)


ansible/playbooks/start-redpanda.yml
```diff
@@ -7,44 +7,63 @@
...
-      - restart redpanda-tuner
-      - restart redpanda
     vars:
       use_public_ips: "{{ advertise_public_ips | d() | bool }}"
     shell: |
       set -e
-      rpk config set cluster_id 'redpanda'
-      rpk config set organization 'redpanda-test'
-      rpk config set redpanda.advertised_kafka_api '{
+      sudo -u redpanda rpk redpanda config set cluster_id 'redpanda'
```
vuldin, line 14: All `rpk` commands are updated to the new syntax to remove the deprecated warnings from the logs. Also the commands are now being ran by the redpanda user (since that is the owner of the files being modified).

aws/cluster.tf
```diff
@@ -82,18 +84,18 @@ resource "aws_security_group" "node_sec_group" {
...
+    self      = true
   }
 
   # HTTP access to the Admin port
   ingress {
     from_port   = 9644
     to_port     = 9644
     protocol    = "tcp"
-    self        = true
+    cidr_blocks = ["0.0.0.0/0"]
```
vuldin, line 98: This opens port 9644 to external traffic, which was causing an issue @r-vasquez and I saw while testing Tristan's PR earlier.

ansible/playbooks/start-redpanda.yml
r-vasquez: Nice, thanks!

aws/vars.tf
```diff
@@ -9,6 +9,12 @@ variable "nodes" {
   default     = "3"
 }
 
+variable "ha" {
+  description = "Whether to use placement groups to create an HA topology"
+  type        = bool
+  default     = false
+}
+
```
r-vasquez, line 17: Can you update the README of `aws/` with this variable?

This can be done in a follow-up/v2 branch since I think the Readme is outdated.

r-vasquez (approved):
This is great, thanks a lot!

aws/vars.tf
vuldin: Went ahead and included this in the most recent commit.

PR #86, opened 2022-10-24
Enable tiered storage in AWS via IAM policy

vuldin:
This PR adds tiered storage feature in AWS deployments. Once merged, new PRs will be created for adding the same feature to other cloud providers. Tiered storage is disabled by default (`tiered_storage_enabled` in `vars.tf`). When tiered storage is enabled, then by default it is enabled across all topics in this PR.

### Steps to run
```
git clone https://github.com/redpanda-data/deployment-automation.git
cd deployment-automation
git checkout enable-si
cd aws
terraform init -upgrade
terraform apply -var "enable_monitoring=false" -var "tiered_storage_enabled=true"
cd ..
ansible-playbook --private-key `cat ~/.ssh/id_rsa.pub | awk '{print $2}'` -i hosts.ini -v ansible/playbooks/provision-node.yml -e '{ "redpanda": { "cluster": { "cloud_storage_segment_max_upload_interval_sec": 30 }}}'
```

Now connect via ssh to a broker, create a topic, produce some data, and check S3 bucket for remote segments. If you need to retrieve the IP address of one of the nodes, run `terraform output`.
```
ssh ubuntu@<redpanda-ec2-ip>
rpk topic create alog
BATCH=$(date) ; printf "$BATCH %s\n" {1..1000} | rpk topic produce alog
```

Run the last command 20+ times. View contents of the bucket in AWS console to see remote segments.


ansible/playbooks/start-redpanda.yml
```diff
@@ -61,3 +61,43 @@
...
+      {% if use_public_ips %}
+      sudo -u redpanda rpk cluster config set cloud_storage_bucket {{ si_bucket_name }} --api-urls {{ "%s:%d" % (inventory_hostname, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_region {{ aws_region }} --api-urls {{ "%s:%d" % (inventory_hostname, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_access_key ABCDEFGHIJKLMNOP --api-urls {{ "%s:%d" % (inventory_hostname, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_secret_key 1234567890abcdefghijklmnop --api-urls {{ "%s:%d" % (inventory_hostname, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_enable_remote_read true --api-urls {{ "%s:%d" % (inventory_hostname, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_enable_remote_write true --api-urls {{ "%s:%d" % (inventory_hostname, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_segment_max_upload_interval_sec 30 --api-urls {{ "%s:%d" % (inventory_hostname, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_credentials_source aws_instance_metadata --api-urls {{ "%s:%d" % (inventory_hostname, 9644) }}
```
vuldin, line 80: These lines may not be needed. We may not want to enable global `cloud_storage_remote_read`/`cloud_storage_remote_write`, `cloud_storage_max_upload_interval_sec` is being set here only to make it easier to detect remote segments, and I was told `cloud_storage_credentials_source` was needed but I have seen this work without it (although I cannot reproduce).

vuldin:
Question: Should shadow indexing (an enterprise feature) be enabled by default in this project? Right now it is.

ansible/playbooks/start-redpanda.yml
```diff
@@ -61,3 +61,43 @@
...
+      sudo -u redpanda rpk cluster config set cloud_storage_enabled true --api-urls {{ "%s:%d" % (inventory_hostname, 9644) }}
+      {% else %}
+      sudo -u redpanda rpk cluster config set cloud_storage_bucket {{ si_bucket_name }} --api-urls {{ "%s:%d" % (private_ip, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_region {{ aws_region }} --api-urls {{ "%s:%d" % (private_ip, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_access_key ABCDEFGHIJKLMNOP --api-urls {{ "%s:%d" % (private_ip, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_secret_key 1234567890abcdefghijklmnop --api-urls {{ "%s:%d" % (private_ip, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_enable_remote_read true --api-urls {{ "%s:%d" % (private_ip, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_enable_remote_write true --api-urls {{ "%s:%d" % (private_ip, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_segment_max_upload_interval_sec 30 --api-urls {{ "%s:%d" % (private_ip, 9644) }}
```
vuldin, line 89: `cloud_storage_segment_max_upload_interval_sec` is being set here just to make it easy to see remote segments without waiting. It may need to be left out.

README.md
```diff
@@ -42,6 +42,7 @@ You can pass the following variables as `-e var=value`:
 * `advertise_public_ips=false|true`: Configure Redpanda to advertise the node's public IPs for client communication instead of private IPs. This allows for using the cluster from outside its subnet. **Note**: This is not recommended for production deployments, because it means that your nodes will be public. Use it for testing only. Default `false`
 * `grafana_admin_pass=<password_here>`: Configure Grafana's admin user's password
 * `ephemeral_disk`: Enable filesystem check for attached disk, useful when using attached disks in instances with ephemeral OS disks (i.e Azure L Series). This allows a filesystem repair at boot time and ensures that the drive is remounted automatically after a reboot. Default `false` 
+* `si_max_upload_interval`: Maps to `cloud_storage_segment_max_upload_interval_sec`. Sets the number of seconds for idle timeout. If this property is empty, Redpanda uploads metadata to the cloud storage, but the segment is not uploaded until it reaches the `segment.bytes` size. By default, the property is empty.
```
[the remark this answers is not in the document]
rkruze: I agree it is a bit verbose, but let's use the same name. I wonder at this point we might want to consider a config file with all the variables listed vs just being documented?

aws/cluster.tf
```diff
@@ -14,15 +15,69 @@ locals {
...
+  name       = "${var.instance_name_prefix}-redpanda"
+  roles      = [aws_iam_role.redpanda.name]
+  policy_arn = aws_iam_policy.redpanda.arn
+}
+
+resource "aws_iam_instance_profile" "redpanda" {
+  name = "${var.instance_name_prefix}-redpanda"
+  role = aws_iam_role.redpanda.name
+}
+
```
rkruze, line 65: Should this all be conditional based on having SI enabled?

aws/vars.tf
```diff
@@ -95,3 +101,9 @@ variable "distro_ssh_user" {
     "amazon-linux-2" = "ec2-user"
   }
 }
+
+variable "si_enabled" {
+  description = "Enables or disables shadow indexing"
+  type = bool
+  default = false
+}
```
rkruze, line 109: Let's call this tiered storage, not SI.

rkruze:
Let's rename this to tiered storage instead of SI. As this is what the docs are focused around.

ansible/playbooks/start-redpanda.yml
```diff
@@ -61,3 +61,32 @@
...
+      - restart redpanda-tuner
+      - restart redpanda
+    vars:
+      use_public_ips: "{{ advertise_public_ips | d() | bool }}"
+      this_node: "{{ inventory_hostname if use_public_ips else private_ip }}"
+    shell: |
+      sudo -u redpanda rpk cluster config set cloud_storage_bucket {{ si_bucket_name }} --api-urls {{ "%s:%d" % (this_node, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_region {{ aws_region }} --api-urls {{ "%s:%d" % (this_node, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_access_key ABCDEFGHIJKLMNOP --api-urls {{ "%s:%d" % (this_node, 9644) }}
+      sudo -u redpanda rpk cluster config set cloud_storage_secret_key 1234567890abcdefghijklmnop --api-urls {{ "%s:%d" % (this_node, 9644) }}
```
r-vasquez, line 76: Shouldn't this `access_key` and `secret_key` be overridden by an ansible var?

ansible/playbooks/start-redpanda.yml
```diff
@@ -61,3 +61,32 @@
...
+
+  - name: enable shadow indexing
+    notify:
+      - restart redpanda-tuner
+      - restart redpanda
+    vars:
+      use_public_ips: "{{ advertise_public_ips | d() | bool }}"
+      this_node: "{{ inventory_hostname if use_public_ips else private_ip }}"
+    shell: |
+      sudo -u redpanda rpk cluster config set cloud_storage_bucket {{ si_bucket_name }} --api-urls {{ "%s:%d" % (this_node, 9644) }}
```
r-vasquez, line 73: I think we can add a var for api_urls that contains all the admin API endpoints, similar to what you did [here](https://github.com/redpanda-data/deployment-automation/commit/8579fdde423714696881c07dc1ee331652c31ffd#diff-c28b7b45769b3c8b11796f623bc3b3aa49a28fcd79d5f0cbb2d6ebc757994e8aR58)

aws/cluster.tf
tmgstevens: +1

aws/vars.tf
```diff
@@ -15,6 +15,12 @@ variable "ha" {
   default     = false
 }
 
+variable "instance_name_prefix" {
```
tmgstevens, line 18: I'm not really sure about the purpose of `instance_name_prefix`. Perhaps we need a way of overriding the `deployment_id` so that it isn't always UUID-timestamp - so if a user wanted to put their project codename or something then they could.

ansible/playbooks/start-redpanda.yml
vuldin: The only reason these two values are here is because they are required by Redpanda, but they are never being used in this implementation. This instead relies on the permissions available from the associated IAM profile.

But I agree that we would need to replace these values with user-provided ones if that approach was being implemented. If we do decide to implement that approach, then I was hoping for that to be in a separate PR so that this one just focuses on the IAM profile approach.

aws/vars.tf
vuldin: I agree that these two variables are too similar. I'll change the name of this variable to `deployment_prefix` and have it be empty by default. Then I'll update the comment to know that that will override `deployment_id` if set, and that `deployment_id` already has a unique default value. I'll also switch all current uses of `instance_name_prefix` to `deployment.id`.

aws/cluster.tf
vuldin: I'll address this in the next push. The user will be able to set `deployment_prefix`, but by default it will be blank. If left blank, then the default `deployment_id` value will be used.

ansible/playbooks/start-redpanda.yml
tmgstevens: I think we should set them to something like "THISVALUENOTUSED" and a comment nearby to explain why.

aws/cluster.tf
```diff
@@ -3,9 +3,10 @@ resource "random_uuid" "cluster" {}
 resource "time_static" "timestamp" {}
 
 locals {
-  uuid          = random_uuid.cluster.result
-  timestamp     = time_static.timestamp.rfc3339
-  deployment_id = "redpanda-${local.uuid}-${local.timestamp}"
+  uuid                       = random_uuid.cluster.result
+  timestamp                  = time_static.timestamp.unix
+  deployment_id              = length(var.deployment_prefix) > 0 ? var.deployment_prefix : "redpanda-${substr(local.uuid, 0, 8)}-${local.timestamp}"
```
vuldin, line 8: Terraform complained about names being too long, so I shortened both the UUID and the timestamp.

aws/provider.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "3.73.0"
+      version = "4.35.0"
```
tmgstevens, line 5: Please update the readme

vuldin:
I'll squash the PR feedback commits once the PR is approved (let me know if you think this is unneeded).

ansible/playbooks/roles/redpanda_broker/templates/configs/tiered_storage.j2
```diff
@@ -0,0 +1,10 @@
+cluster:
+  cloud_storage_access_key: THISVALUENOTUSED
+  cloud_storage_bucket: {{ tiered_storage_bucket_name if tiered_storage_bucket_name is defined }}
+  cloud_storage_enable_remote_read: true
+  cloud_storage_enable_remote_write: true
```
vuldin, line 5: This makes all topics tiered-storage-enabled by default if the `tiered_storage_enabled` terraform variable is enabled.

ansible/playbooks/roles/redpanda_broker/vars/main.yml
```diff
@@ -3,4 +3,6 @@
 custom_config_templates:
   - template: configs/defaults.j2
   - template: configs/tls.j2
-    condition: "{{ tls | default(False) | bool }}"
\ No newline at end of file
+    condition: "{{ tls | default(False) | bool }}"
+  - template: configs/tiered_storage.j2
+    condition: "{{ tiered_storage_bucket_name is defined | default(False) | bool }}"
```
vuldin, line 8: The `tiered_storage_bucket_name` ansible variable is pulled from `hosts.ini`, and is only defined if the `tiered_storage_enabled` terraform variable is `true`.

aws/cluster.tf
```diff
@@ -14,15 +15,73 @@ locals {
...
+  name   = local.deployment_id
+  path   = "/"
+  policy = jsonencode({
+    Version   = "2012-10-17"
+    Statement = [
+      {
+        "Effect": "Allow",
+        "Action": [
+          "s3:*",
+          "s3-object-lambda:*",
```
vuldin, line 29: This needs to be more limited to list only those needed permissions.

vuldin: Discussed this with the broader team. From that conversation:
> Having an policy that tightly limits redpanda to its own bucket is clearly the right thing to do, it's less obvious to me that restricting the verbs is helpful: it creates maintenance burden to keep those up to date as/when we change redpanda, and doesn't meaningfully change security if redpanda is the owner of the bucket.

So limiting this policy to the specific bucket (as shown in this file below) is the best approach, and avoids a maintenance burden as tiered storage capabilities are expanded.

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
```diff
@@ -10,7 +10,7 @@ node:
     - address: {{ hostvars[inventory_hostname].advertised_ip }}
       port: {{ redpanda_kafka_port }}
     advertised_rpc_api:
-    - address: {{ hostvars[inventory_hostname].advertised_ip }}
+      address: {{ hostvars[inventory_hostname].advertised_ip }}
```
vuldin, line 13: This resulted in an error due to the fact that `advertised_rpc_api` (unlike other listeners) isn't a list.

ansible/playbooks/roles/redpanda_broker/templates/configs/tiered_storage.j2
```diff
@@ -0,0 +1,10 @@
...
+  cloud_storage_region: {{ aws_region if aws_region is defined }}
+  cloud_storage_secret_key: THISVALUENOTUSED
+  cloud_storage_credentials_source: aws_instance_metadata
```
tmgstevens, line 8: Need to pass this in as a param from terraform otherwise we're assuming AWS only

hcoyote:
Temporarily cherry picked the ansible2.14+ node_exporter fixes into my local repo in order to get this to run. fwiw, it seems to run ok with the ansible 7.1.0 bottle in homebrew (or, at least, it's behaving the same way so far as other clusters I've run this week from a cluster provisioning standpoint).

aws/cluster.tf
hcoyote: how safe is it to just grab the first 8 characters off the uuid? I haven't look at how go-uuid works to see if it's one of the time-based uuids.

aws/s3.tf
```diff
@@ -0,0 +1,19 @@
...
+  acl    = "private"
+}
+
+resource "aws_s3_bucket_versioning" "tiered_storage" {
+  count  = var.tiered_storage_enabled ? 1 : 0
+  bucket = aws_s3_bucket.tiered_storage[count.index].id
+  versioning_configuration {
+    status = "Disabled"
+  }
+}
```
hcoyote, line 19: Can we get a comment added on here on why we disable versioning (and maybe what happens, good or bad, if the user overrides this)?

hcoyote (approved):
added some questions for clarity, I don't see anything specific that should hold this back as is.

ansible/playbooks/roles/redpanda_broker/templates/configs/tiered_storage.j2
```diff
@@ -0,0 +1,10 @@
+cluster:
+  cloud_storage_access_key: THISVALUENOTUSED
```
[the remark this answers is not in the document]
vuldin: No, this is just a placeholder. Unfortunately Redpanda requires this to be something even when we are using `aws_instance_metadata` (IAM permissions applied to the EC2 instance).

tmgstevens:
Thanks for the review @hcoyote

PR #90, opened 2022-11-02
Extensible configuration

tmgstevens:


vuldin (commented):
I really like this layered config file approach. It will make passing in variables much simpler. I'm still not quite sure how to modify parameter values from a flag on the command, but I will do another review once the README is updated and I'm able to test this part out.

vuldin (approved):
LGTM. Tested modifying node and cluster-level parameters from passing flag on command line and everything works as expected.

PR #93, opened 2022-11-09
Introduce logic to support use of unstable versions

tmgstevens:
Allow use of staging repo and specification of specific redpanda versions.
Updated README to include details of defaults


vuldin (approved):
LGTM

PR #95, opened 2022-11-09
adopt new cluster boostrapping guidance

andrwng:
Starting with v22.3, Redpanda can automatically assign node IDs to empty nodes as they join the cluster, and there is no longer a requirement to configure a single root node to start the cluster.

This commit updates the templates and tasks to use seeds-driven bootstrap, where all seed servers are expected to be configured with an identical list of seed servers. Once all seed servers are up they verify with each other that they all agree on the seeds, and collectively start a cluster with all seed servers as the initial set of nodes in the controller Raft group.

Admittedly, this commit doesn't do anything to ensure that the installed version of Redpanda actually supports the new mode. In the short term, we can address versioning with documentation and project branching.


tmgstevens:
@andrwng do you think we should be setting the seed servers in the redpanda.yml file?

andrwng:
> @andrwng do you think we should be setting the seed servers in the redpanda.yml file?

Yeah that's a good point. Frankly, the `bootstrap` command is a bit redundant now that we have identical configs that are known statically. I updated the PR to reflect that.

I haven't tested this though. Is there a way to test using nightlies?

tmgstevens:
> I haven't tested this though. Is there a way to test using nightlies?

If you check out #93 we've got the ability to specify either the unstable repo, or you can point to a custom repo instead.

tmgstevens:
Let's add a check as follows:
```
- name: Check version compatibility
  ansible.builtin.assert:
    that:
      - redpanda_version == 'latest' or redpanda_version is version('22.3', '>=')
    fail_msg: "This version of the Redpanda role is only supported with Redpanda 22.3 or later"
    success_msg: "'redpanda_version' is 22.3 or later"
```

andrwng:
> > I haven't tested this though. Is there a way to test using nightlies?
> 
> If you check out #93 we've got the ability to specify either the unstable repo, or you can point to a custom repo instead.

Actually looking at the defaults, I think I'll keep the `bootstrap` command in. It does add some configs that aren't there: Kafka APIs, etc.

andrwng:
> Let's add a check as follows:
> 
> ```
> - name: Check version compatibility
>   ansible.builtin.assert:
>     that:
>       - redpanda_version == 'latest' or redpanda_version is version('22.3', '>=')
>     fail_msg: "This version of the Redpanda role is only supported with Redpanda 22.3 or later"
>     success_msg: "'redpanda_version' is 22.3 or later"
> ```

Sounds good, where should this go?

tmgstevens:
> Sounds good, where should this go?

Probably the top of `install_redpanda.yml`, before we actually do any damage!

PR #98, opened 2022-12-02
ansible: fix advertised apis

andrwng:
We included two settings for both advertised_*_apis, one of which incorrectly pointed at the private IP rather than the advertised IP.


r-vasquez (approved):
Thanks!

tmgstevens (approved):
Looks like a merge error

PR #105, opened 2023-01-12
Bugfix/issue 103/shell cmd warn

hcoyote:
ansible 2.14+ doesn't support warn parameters for ansible.builtin.shell and causes playbook runs to fail.

This fixups the code where we use it explicitly. Complicating this issue, is the fact that the cloudalchemy.node_exporter module also has the same issue, but is no longer maintained. I've switched us over to geerlingguy's fork which is more recently updated and fixes the underlying problem.


tmgstevens (approved):
LGTM, thanks for your first commit @hcoyote !

PR #107, opened 2023-01-12
fix: add in comments and context on why the playbook tasks work in pa…

hcoyote:
…rticular ways

### DIFF
--- a/ansible/playbooks/roles/redpanda_broker/tasks/start-redpanda.yml
+++ b/ansible/playbooks/roles/redpanda_broker/tasks/start-redpanda.yml
@@ -171,6 +171,9 @@
 # once the cluster is initialized, this handles subsequent config update operations
 # for future ansible runs.
 #
+# changes to the on-disk config at this point will trigger a safe restart to occur
+# with ansible if restart handling is configured in the role settings.
+#
 - name: Generate Node config - post-bootstrap runs
   ansible.builtin.template:
     src: "redpanda.yml"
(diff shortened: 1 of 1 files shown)


ansible/playbooks/roles/redpanda_broker/tasks/start-redpanda.yml
```diff
@@ -83,13 +147,20 @@
...
 - name: Check if restart needed
   ansible.builtin.shell:
     cmd: "rpk cluster config status | grep '^{{ node_id }} ' | awk '{ print $3 }' |  grep -E 'true|false'"
   register: restart_required_rc
   changed_when: False
 
-- name: Generate Node config
+# We need to run the node config again after first run because some things
+# could have changed out-of-band. Since rpk can modify the config on your
+# behalf, we want to reset it back to a known templated state.
```
tmgstevens, line 161: This task will only fire if the playbook is being run on an already initialised cluster, for example, when running the TLS playbooks. So here we're going to set any node properties that we or the user have specified and this will take effect on the next restart.
We can't do it sooner could it could potentially have caused the `rpk cluster config` or `rpk cluster status` commands to have failed

tmgstevens:
One minor comment and then I'm happy to merge this.

ansible/playbooks/roles/redpanda_broker/tasks/start-redpanda.yml
hcoyote: Ok, I'll integrate that into the comment. Still seems weird to define this twice even though it's basically doing the same thing. Seems like we'd be better off using a notify on those tasks that should trigger a rebuild of the config, and then stick the config build task in a handler.

hcoyote: Ok. I updated comment to reflect the above.

tmgstevens (approved):

PR #113, opened 2023-01-20
Fix build errors on GCP and Azure with tiered storage.

tmgstevens:
Builds on Azure and GCP failing with missing variables.

### DIFF
--- a/azure/outputs.tf
+++ b/azure/outputs.tf
@@ -33,15 +33,18 @@ output "public_key_path" {
 resource "local_file" "hosts_ini" {
   content = templatefile("${path.module}/../templates/hosts_ini.tpl",
     {
-      redpanda_public_ips  = "${azurerm_public_ip.redpanda.*.ip_address}"
-      redpanda_private_ips = "${azurerm_network_interface.redpanda.*.private_ip_address}"
-      client_public_ips    = "${azurerm_public_ip.redpanda_client.*.ip_address}"
-      client_private_ips   = "${azurerm_network_interface.redpanda_client.*.private_ip_address}"
-      monitor_public_ip    = var.enable_monitoring ? "${azurerm_public_ip.monitoring.0.ip_address}" : ""
-      monitor_private_ip   = var.enable_monitoring ? "${azurerm_network_interface.monitoring.0.private_ip_address}" : ""
-      enable_monitoring    = "${var.enable_monitoring}"
-      ssh_user             = "${var.admin_username}"
-      rack                 = "${azurerm_linux_virtual_machine.redpanda.*.platform_fault_domain}"
+      redpanda_public_ips        = "${azurerm_public_ip.redpanda.*.ip_address}"
+      redpanda_private_ips       = "${azurerm_network_interface.redpanda.*.private_ip_address}"
+      client_public_ips          = "${azurerm_public_ip.redpanda_client.*.ip_address}"
+      client_private_ips         = "${azurerm_network_interface.redpanda_client.*.private_ip_address}"
+      monitor_public_ip          = var.enable_monitoring ? "${azurerm_public_ip.monitoring.0.ip_address}" : ""
+      monitor_private_ip         = var.enable_monitoring ? "${azurerm_network_interface.monitoring.0.private_ip_address}" : ""
+      enable_monitoring          = "${var.enable_monitoring}"
+      ssh_user                   = "${var.admin_username}"
+      rack                       = "${azurerm_linux_virtual_machine.redpanda.*.platform_fault_domain}"
+      cloud_storage_region       = var.region
+      tiered_storage_enabled     = false
+      tiered_storage_bucket_name = ""
     }
   )
   filename = "${path.module}/../hosts.ini"
--- a/gcp/cluster.tf
+++ b/gcp/cluster.tf
@@ -18,12 +18,6 @@ resource "google_compute_resource_policy" "redpanda-rp" {
     availability_domain_count = var.ha ? max(3, var.nodes) : 1
   }
   count = var.ha ? 1 : 0
-  lifecycle {
-    precondition {
-      condition = var.ha && var.nodes <= 8
-      error_message = "GCP does not support resource policies with greater than 8 nodes"
-    }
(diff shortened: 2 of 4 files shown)


gcp/cluster.tf
```diff
@@ -18,12 +18,6 @@ resource "google_compute_resource_policy" "redpanda-rp" {
     availability_domain_count = var.ha ? max(3, var.nodes) : 1
   }
   count = var.ha ? 1 : 0
-  lifecycle {
-    precondition {
-      condition = var.ha && var.nodes <= 8
-      error_message = "GCP does not support resource policies with greater than 8 nodes"
-    }
-  }
```
hcoyote, line 26: Why was this removed? Not sure I understand how this is related to making the region selection work generically for all three clouds.

azure/outputs.tf
```diff
@@ -33,15 +33,18 @@ output "public_key_path" {
...
+      client_public_ips          = "${azurerm_public_ip.redpanda_client.*.ip_address}"
+      client_private_ips         = "${azurerm_network_interface.redpanda_client.*.private_ip_address}"
+      monitor_public_ip          = var.enable_monitoring ? "${azurerm_public_ip.monitoring.0.ip_address}" : ""
+      monitor_private_ip         = var.enable_monitoring ? "${azurerm_network_interface.monitoring.0.private_ip_address}" : ""
+      enable_monitoring          = "${var.enable_monitoring}"
+      ssh_user                   = "${var.admin_username}"
+      rack                       = "${azurerm_linux_virtual_machine.redpanda.*.platform_fault_domain}"
+      cloud_storage_region       = var.region
+      tiered_storage_enabled     = false
+      tiered_storage_bucket_name = ""
```
hcoyote, line 47: nitpick: we should be careful about whitespace-only changes included in fix diffs, makes it a little harder to review (yes, I know we can configure PR reviews to ignore whitespace 😄 )

tmgstevens: It's not whitespace only, the last three lines are relevant

gcp/cluster.tf
tmgstevens: It's not, but it doesn't work with this condition in. We could do it as separate PRs, but figured a single catch-all "make stuff work" PR would be fine

azure/outputs.tf
hcoyote: Understood.

gcp/cluster.tf
hcoyote: Got it.

tmgstevens: I'd like to understand why it suddenly stopped working though

azure/outputs.tf
```diff
@@ -33,15 +33,18 @@ output "public_key_path" {
...
     }
```
hcoyote, line 48: this looks fine, but we should probably have this whole local_file resource in cluster.tf like the other three clouds.

templates/hosts_ini.tpl
```diff
@@ -1,14 +1,14 @@
...
+${ ip } ansible_user=${ ssh_user } ansible_become=True private_ip=${redpanda_private_ips[i]}%{ if rack[i] != null } rack=${rack[i]}%{ endif }%{ if tiered_storage_enabled } tiered_storage_bucket_name=${tiered_storage_bucket_name} cloud_storage_region=${cloud_storage_region}%{ endif }
 %{ endfor ~}
 %{ if enable_monitoring }
 
 [monitor]
 ${ monitor_public_ip } ansible_user=${ ssh_user } ansible_become=True private_ip=${ monitor_private_ip }
 %{ endif }
-%{ if client_count > 0 }
 
+%{ if length(client_public_ips) > 0 }
```
hcoyote, line 11: this will end up a noop if the user adds clients with no public ips, right? or is the TF only capable of building clients with public front ends right now?

gcp/cluster.tf
hcoyote: Let's file an issue on this one in particular. after reading the google_compute_resource_policy flow in clusters.tf and looking at the docs, I don't understand where this limitation of 8 nodes comes from

templates/hosts_ini.tpl
tmgstevens: My understanding is that this would be an array of nulls if the clients don't have public ip addresses. Aiming for a noop if there are no clients.

gcp/cluster.tf
tmgstevens: The issue here is that GCP does not give you visibility nor control over which failure domain a resource has been placed into (https://issuetracker.google.com/issues/256993209?pli=1). So the only way that we can guarantee that specific nodes are in separate racks is to put them into entirely separate failure domains - basically one broker per failure domain, and we are limited by the number of failure domains (at the moment 8).

hcoyote: Got it! This explanation would be worth adding as a comment.

templates/hosts_ini.tpl
tmgstevens: In AWS we have: 
```
output "client" {
  value = {
    for instance in aws_instance.client :
    instance.public_ip => instance.private_ip
  }
}
```
In Azure:
```
output "client_public_ips" {
  value = "${azurerm_public_ip.redpanda_client.*.ip_address}"
}
```
And GCP:
```
google_compute_instance.client.*.network_interface.0.access_config.0.nat_ip
```

I think this is okay, especially given the next line:
```
%{ for i, ip in client_public_ips ~}
${ ip } ansible_user=${ ssh_user } ansible_become=True private_ip=${client_private_ips[i]} id=${i}
%{ endfor ~}
```

tmgstevens: I've just re-tested this on all three. We actually don't give the user an option to configure these without a public IP address. I'm sure they could change it to do so, but also bear in mind that public doesn't have to mean _public_, it could still be a subnet based IP address.
I think we're good here.

hcoyote (approved):
I'm not able to test this on gcp, but it's definitely working on azure. The changes look reasonable and we can pickup any additional stuff in future PRs.

PR #118, opened 2023-01-27
Implement more sophisticated serialization of the safe restart method…

### LINKED ISSUE #115: Ansible fails when enabling TLS due to safe restart
The code for safe restart uses `rpk cluster maintenance enable {{ node_id }}` which unfortunately uses the new `rpk` settings before TLS is enabled.

We either need a way to single-thread the whole of the last few plays (to ensure that maintenance mode is enabled, the node is restarted and then MM disabled all in a single action), or we need to nestle the updating of the `redpanda.yaml` file into that single shell command.

When I wrote this I couldn't find a way of having a block with throttle set within a role.

tmgstevens:
… that allows us to bring in the modification of /etc/redpanda/redpanda.yaml into the part after we've put the node into maintenance mode.

Fixes #115

### DIFF
--- a/ansible/playbooks/roles/redpanda_broker/tasks/safe-restart.yml
+++ b/ansible/playbooks/roles/redpanda_broker/tasks/safe-restart.yml
@@ -3,11 +3,11 @@
 # only one node out of service at a time.
 #
 - name: Enable Maintenance Mode
-  ansible.builtin.shell:
+  ansible.builtin.command:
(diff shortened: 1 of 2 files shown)


hcoyote (approved):
Tested. Seems like it's doing the right thing to me from the rolling restart perspective.

hcoyote:
We'll want to revisit this once auth support comes into the playbook, as this may have repercussions for how we interact with rpk, especially when using mTLS per @jcsp.

Issue #81, opened 2022-10-17
Switch to /public_metrics endpoint


vuldin:
The grafana dashboard is generated based off the internal `/metrics` endpoint. This should be switched to the public-facing `/public_metrics` endpoint.

https://github.com/redpanda-data/deployment-automation/blob/01d1c3fa8313a84ba5cb9a1f195161251889ac96/ansible/playbooks/deploy-prometheus-grafana.yml#L5

hcoyote:
I was reading up on this. It looks like the public_metrics endpoint exists on 22.2+. Do we need to special-case this somehow to deal with ansible users sitting at a lower release version?

https://docs.redpanda.com/docs/manage/monitoring/

Also, is there anything more to do here than to update `--metrics-endpoint` to use the `/public_metrics` endpoint? Should it still use the private_ip for the host or should it work off a public_ip?

hcoyote:
This should probably instead use the grafana dashboards we publish to the Grafana marketplace instead.

hcoyote:
Alternatively, we should look at enabling releases on the observability project and pull bundled downloads directly from https://github.com/redpanda-data/observability

gene-redpanda:
<img width="509" alt="Screenshot 2023-04-21 at 12 17 01 PM" src="https://user-images.githubusercontent.com/123959009/233685558-1af0a9bd-0120-48d8-82f4-126491fc232e.png">

This is done

PR #125, opened 2023-02-01
Dashboard should only rebuild if missing or new RP version

hcoyote:
To reduce amount of ansible churn, the shell command that generates the grafana dashboard json file should only fire off if the dashboard file is missing or if we get triggered to install a new version of redpanda.


gene-redpanda (approved):

gene-redpanda:
Sensible stuff

hcoyote:
before I merge this, I'm going to do one more test deploying a slightly older RP release and then have the ansible upgrade. I'm 99% sure this works right as is, but want to double check (especially in light of the weirdness I've experienced with package_result in other tests)

tmgstevens:
Someone who knows more about Ansible than I do: Can we rely on `package_result` being set? It's defined in a role further up in the playbook and to that extent is an internal fact to that role. Are we okay relying on that?

hcoyote:
> Registered variables are stored in memory. You cannot cache registered variables for use in future playbook runs. Registered variables are only valid on the host for the rest of the current playbook run, including subsequent plays within the same playbook run.

This implies to me that we're good as long as we're not somehow skipping roles/redpanda_broker/tasks/install-redpanda.yml (e.g., setting a tag on it and excluding that tag from runs?)

tmgstevens:
@hcoyote I think now that we have the Grafana dashboards available here: https://grafana.com/grafana/dashboards/?search=redpanda we can avoid the need to generate the dashboards and simply add the relevant IDs in as follows:

https://github.com/cloudalchemy/ansible-grafana#:~:text=Dashboard%20example%3A,datasource%3A%20prometheus

hcoyote:
@tmgstevens yeah, Josh and I were discussing this in relation to #81 earlier this week. It should be tracked there.

hcoyote:
Oh, I didn't realize this was never merged in. 

Let's goahead and merge it and work on doing it the other way in a new commit. There's some edge cases we have to consider when pulling the dashboards from Grafana (e.g., it assumes you're internet connected, for example).

PR #126, opened 2023-02-01
Bugfix/issue 120/nodexporter version pin

hcoyote:
Prevent geerlinguy.node_exporter from unnecessarily pinging github on every run of ansible from every node. Pinning to a particular version for now.


ansible/playbooks/install-node-deps.yml
```diff
@@ -1,5 +1,7 @@
 ---
 - hosts: redpanda
+  vars: 
+    node_exporter_version: 1.5.0
```
tmgstevens, line 4: Can we make this into `node_exporter_version: node_exporter_version | default('1.5.0')` please? This allows the user to override it if they really need to.

gene-redpanda (approved):

PR #127, opened 2023-02-01
gcp cleanup and labeling

gene-redpanda:
Added labels for all GCP instances to ensure that it is straightforward for users to add tags for inventory management and firewall purposes. Also cleaned up some of the whitespace and removed the tf0.11 interpolation format.


hcoyote (approved):
LGTM

PR #131, opened 2023-02-02
Add docs for performing upgrades and introduce the ability to specify…

tmgstevens:
… options for `rpk` (for when admin authentication is turned on)

@rkruze commented that upgrades weren't documented. Included new parameter per request of @jcsp to allow for upgrading of clusters with SASL auth turned on for admin interface.

Tested upgrades with and without SASL.


README.md
```diff
@@ -155,6 +156,32 @@ A similar process can be used to build a cluster with TLS in one execution as to
   3. `install-certs.yml` - Installs the certificate and also applies the `redpanda_broker` role to the cluster nodes. Note: This will install and start Redpanda (and restart any brokers that do not have `skip_node=true` set)
 5. If `install-certs.yml` was not run in step iii above, you will need to run `provision-node.yml` which will install the `redpanda_broker` role. **Note: If TLS is enabled on the cluster, make sure that `-e tls=true` is set, otherwise this playbook will disable TLS across any nodes that don't have `skip_nodes=true` set.**
 
+## Upgrading a cluster
```
hcoyote, line 159: how does this handle a downgrade? I'm assuming we depend upon redpanda itself to do the right thing if the version goes lower. But, is there some sort of gating we should do within the ansible to prevent going too far back?

gene-redpanda (approved):

README.md
gene-redpanda: I think the tldr is that we don't really handle a downgrade ATM. It may work (surprised if it did) but it should be considered unsupported

tmgstevens:
> Reasonable as a v1. I think a next step would be to figure out how to integrate ansible-vault in here so we can refer to credentials safely.
> 
> for example: https://www.redhat.com/sysadmin/ansible-playbooks-secrets

@hcoyote at the moment we don't handle authentication at all in the playbook. I'll raise an issue for this. There's no reason you couldn't put `redpanda_rpk_opts` into vault if you wanted to - I can craft an example for that.

README.md
hcoyote: So, looks like we might be able to do this with the package_facts module, a version test to compare, and assert to fail the play.
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_facts_module.html https://docs.ansible.com/ansible/latest/collections/ansible/builtin/version_test.html
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/assert_module.html

PR #134, opened 2023-02-02
various minor tweaks/changes to aws stuff

gene-redpanda:
Added tag merging to support custom tags alongside our existing tags, fixed various formatting issues, cleaned up some deprecated formatting.

Added the GCP splat fixes in too since I missed them before.

### DIFF
--- a/aws/vars.tf
+++ b/aws/vars.tf
@@ -1,4 +1,5 @@
 variable "aws_region" {
+  type        = string
   description = "The AWS region to deploy the infrastructure on"
   default     = "us-west-2"
 }
@@ -10,11 +11,13 @@ variable "clients" {
 }
 
 variable "client_distro" {
+  type        = string
   description = "Linux distribution to use for clients."
   default     = "ubuntu-focal"
 }
 
 variable "client_instance_type" {
+  type        = string
   description = "Default client instance type to create"
   default     = "m5n.2xlarge"
 }
@@ -26,6 +29,7 @@ variable "deployment_prefix" {
 }
 
 variable "distro" {
+  type        = string
   description = "The default distribution to base the cluster on"
   default     = "ubuntu-focal"
 }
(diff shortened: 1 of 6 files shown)


aws/vars.tf
```diff
@@ -194,3 +199,15 @@ variable "tiered_storage_enabled" {
   type        = bool
   default     = false
 }
+
+variable "private_key_path" {
+  type        = string
+  description = "The contents of an SSH key to use for the connection. These can be loaded from a file on disk using the file function. This takes preference over password if provided."
+}
```
hcoyote, line 206: Is this going to force people to set a private_key_path or will the lack of a default here allow this to fall through to the running ssh agent? I guess right now it's implicitly looking for one directly and we're just not setting the variable in a place that people can override easily?

aws/vars.tf
```diff
@@ -143,29 +143,34 @@ variable "public_key_path" {
...
-    }
+  most_recent = true
+
+  filter {
+    name   = "name"
+    values = [
+      "ubuntu/images/hvm-ssd/ubuntu-*-amd64-server-*", "Fedora-Cloud-Base-*.x86_64-hvm-us-west-2-gp2-0",
+      "debian-*-amd64-*", "debian-*-hvm-x86_64-gp2-*'", "amzn2-ami-hvm-2.0.*-x86_64-gp2",
+      "RHEL*HVM-*-x86_64*Hourly2-GP2"
+    ]
```
hcoyote, line 154: For readability, can we do the image name patterns as 1 per line?

hcoyote (changes requested):

aws/vars.tf
gene-redpanda: Well as far as I can tell we just didn't have a variable for private key. So it threw an error on plan. 

I'm fine making that element optional, but I'm pretty sure it's needed for the ansible connection.

aws/vars.tf
gene-redpanda: Sure, nbd

aws/vars.tf
hcoyote: We only need to load the public key to make ansible work. I'm only loading a pub key on my builds and relying upon an agent.

gene-redpanda: defaults to null

hcoyote (approved):

PR #136, opened 2023-02-15
various changes to support running tls

gene-redpanda:
supports running in configurations
* internal TLS only
* internal and external TLS
* no TLS

I've tested it pretty rigorously, but will be running the full testing process again to guarantee everything still works just in case. 

I've included bootstrap scripts in the bootstrap directory to provide quick standup and teardown and even some shell to derive REDPANDA_BROKERS from the hosts.ini to speed testing + do some testing here and there.

edit:

Made some changes to simplify things after realizing something that in retrospect is pretty obvious 😅

### DIFF
--- a/ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
+++ b/ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
@@ -49,8 +49,8 @@ node:
     seed_servers:
 {% for host in groups["redpanda"] %}
       - host:
-        address: {{ hostvars[host]['private_ip'] }}
-        port: {{ redpanda_rpc_port }}
+          address: {{ hostvars[host]['private_ip'] }}
+          port: {{ redpanda_rpc_port }}
 {% endfor %}
 {% if rack is defined %}
     rack: "{{ rack | default('null') }}"
--- a/ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
+++ b/ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
@@ -7,20 +7,8 @@ node:
   redpanda:
     empty_seed_starts_cluster: false
     advertised_kafka_api:
-    - address: {{ hostvars[inventory_hostname].private_ip }}
-      port: {{ redpanda_kafka_port }}
-{% if advertise_public_ips is defined %}
-{% if advertise_public_ips %}
-      name: internal
-{% endif %}
-{% endif %}
-{% if advertise_public_ips is defined %}
-{% if advertise_public_ips %}
     - address: {{ hostvars[inventory_hostname].advertised_ip }}
-      port: {{ redpanda_external_kafka_port }}
-      name: external
-{% endif %}
-{% endif %}
+      port: {{ redpanda_kafka_port }}
     advertised_rpc_api:
       address: {{ hostvars[inventory_hostname].private_ip }}
       port: {{ redpanda_rpc_port }}
@@ -29,20 +17,8 @@ node:
       address: {{ hostvars[inventory_hostname].private_ip }}
       port: {{ redpanda_rpc_port }}
     kafka_api:
-    - address: {{ hostvars[inventory_hostname].private_ip }}
-      port: {{ redpanda_kafka_port }}
-{% if advertise_public_ips is defined %}
-{% if advertise_public_ips %}
-      name: internal
-{% endif %}
-{% endif %}
-{% if advertise_public_ips is defined %}
-{% if advertise_public_ips %}
     - address: 0.0.0.0
-      port: {{ redpanda_external_kafka_port }}
-      name: external
-{% endif %}
-{% endif %}
+      port: {{ redpanda_kafka_port }}
     admin:
     - address: {{ hostvars[inventory_hostname].private_ip }}
       port: {{ redpanda_admin_api_port }}
--- a/ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
+++ b/ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
@@ -17,7 +17,7 @@ node:
       address: {{ hostvars[inventory_hostname].private_ip }}
       port: {{ redpanda_rpc_port }}
     kafka_api:
-    - address: 0.0.0.0
+    - address: {{ hostvars[inventory_hostname].private_ip }}
       port: {{ redpanda_kafka_port }}
     admin:
     - address: {{ hostvars[inventory_hostname].private_ip }}
--- a/ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
+++ b/ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
@@ -7,10 +7,22 @@ node:
   redpanda:
     empty_seed_starts_cluster: false
     advertised_kafka_api:
-    - address: {{ hostvars[inventory_hostname].advertised_ip }}
+    - address: {{ hostvars[inventory_hostname].private_ip }}
       port: {{ redpanda_kafka_port }}
+{% if advertise_public_ips is defined %}
+{% if advertise_public_ips %}
+      name: internal
+{% endif %}
+{% endif %}
+{% if advertise_public_ips is defined %}
+{% if advertise_public_ips %}
+    - address: {{ hostvars[inventory_hostname].advertised_ip }}
+      port: {{ redpanda_external_kafka_port }}
+      name: external
+{% endif %}
+{% endif %}
     advertised_rpc_api:
-      address: {{ hostvars[inventory_hostname].advertised_ip }}
+      address: {{ hostvars[inventory_hostname].private_ip }}
       port: {{ redpanda_rpc_port }}
     data_directory: "{{ redpanda_data_directory }}"
     rpc_server:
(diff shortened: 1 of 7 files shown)


ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
```diff
@@ -7,10 +7,22 @@ node:
   redpanda:
     empty_seed_starts_cluster: false
     advertised_kafka_api:
-    - address: {{ hostvars[inventory_hostname].advertised_ip }}
```
gene-redpanda, line 10: when only advertised ip is used, the cluster cannot talk amongst its members using their private ips and fails to start

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
```diff
@@ -19,13 +31,25 @@ node:
...
+      name: external
+{% endif %}
+{% endif %}
     admin:
     - address: {{ hostvars[inventory_hostname].private_ip }}
       port: {{ redpanda_admin_api_port }}
     seed_servers:
 {% for host in groups["redpanda"] %}
       - host:
-        address: {{ hostvars[host]['inventory_hostname' if (hostvars[host].advertise_public_ips | d() | bool) else 'private_ip'] }}
```
gene-redpanda, line 28: seed servers does not seem to allow for public ip here

r-vasquez (approved):
Tested the scripts with and without TLS :+1:

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
```diff
@@ -10,23 +10,23 @@ node:
...
-        port: {{ redpanda_rpc_port }}
+          address: {{ hostvars[host]['private_ip'] }}
+          port: {{ redpanda_rpc_port }}
```
gene-redpanda, line 29: indent is needed to avoid issues with yaml parsing in rpk/redpanda

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
```diff
@@ -35,12 +35,12 @@ node:
     kafka_api:
       brokers:
 {% for host in groups["redpanda"] %}
-      - {{ hostvars[host]['inventory_hostname' if (hostvars[host].advertise_public_ips | d() | bool) else 'private_ip'] }}:{{ redpanda_kafka_port }}
+      - {{ hostvars[host]['private_ip'] }}:{{ redpanda_kafka_port }}
```
gene-redpanda, line 38: cannot advertise public ip here

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
```diff
@@ -10,23 +10,23 @@ node:
     - address: {{ hostvars[inventory_hostname].advertised_ip }}
       port: {{ redpanda_kafka_port }}
     advertised_rpc_api:
-      address: {{ hostvars[inventory_hostname].advertised_ip }}
+      address: {{ hostvars[inventory_hostname].private_ip }}
```
gene-redpanda, line 13: this must always be private or the cluster cannot talk to itself correctly

tmgstevens: This makes sense

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
tmgstevens: +1

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
```diff
@@ -10,23 +10,23 @@ node:
...
       port: {{ redpanda_rpc_port }}
     data_directory: "{{ redpanda_data_directory }}"
     rpc_server:
       address: {{ hostvars[inventory_hostname].private_ip }}
       port: {{ redpanda_rpc_port }}
     kafka_api:
-    - address: {{ hostvars[inventory_hostname].private_ip }}
+    - address: 0.0.0.0
```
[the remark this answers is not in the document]
tmgstevens: I don't think we need two listeners. Certainly in AWS, `private_ip` can be bound to and it'll automagically map to the EIP if one is presented.

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
tmgstevens: -1
Why wouldn't this be possible? In AWS this is how you need to do things for public IPs.

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
```diff
@@ -35,12 +35,12 @@ node:
...
       brokers:
 {% for host in groups["redpanda"] %}
-      - {{ hostvars[host]['inventory_hostname' if (hostvars[host].advertise_public_ips | d() | bool) else 'private_ip'] }}:{{ redpanda_kafka_port }}
+      - {{ hostvars[host]['private_ip'] }}:{{ redpanda_kafka_port }}
 {% endfor %}
     admin_api:
       addresses:
 {% for host in groups["redpanda"] %}
-      - {{ hostvars[host]['inventory_hostname' if (hostvars[host].advertise_public_ips | d() | bool) else 'private_ip'] }}:{{ redpanda_admin_api_port }}
+      - {{ hostvars[host]['private_ip'] }}:{{ redpanda_admin_api_port }}
```
tmgstevens, line 43: -1, same comment as above

tmgstevens (changes requested):
As far as I can see you're stripping out the ability to work with public_ips. I don't believe that is desirable, but I'll re-test this this afternoon.

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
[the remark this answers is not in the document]
gene-redpanda: if advertise_public_ip is true and Kafka api is set to the private IP while advertised Kafka IP is set to the public IP the rpk client will not be able to communicate with the Kafka api and you will only be able to run rpk commands on a broker.

This change allows us to support rpk commands inside and outside the VPC without regard to the address of the advertised IP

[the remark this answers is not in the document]
tmgstevens: This isn't how it works in AWS, but I'm willing to stand to be corrected in GCP/Azure. In AWS, you can bind to the private IP and it'll automatically map that to the public_ip.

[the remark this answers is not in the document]
gene-redpanda: That was not my experience in testing this past week in AWS unfortunately. I'm willing to check again when I'm back in office on Tuesday just to be sure.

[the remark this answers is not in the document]
tmgstevens: I did just test this and it worked fine for me, but I did need to make your other changes around the rpc and seed addresses. Let's catch up next week. Enjoy your long weekend

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
gene-redpanda: After testing in AWS, using a public IP here causes rpk clients on the broker to use the public IP rather than the private one when communicating to the Redpanda process on the same broker while advertising public ip. While this can work, it seems odd to route traffic on a node outside the network and then back in.

This section is only used for rpk on the broker (per Travis B.) so setting public ip was not useful for other purposes.

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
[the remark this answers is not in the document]
gene-redpanda: That could account for our different experience. I haven't tested private IP in Kafka api while advertising public ip after making the other changes. 

If it does indeed work the same (and I will need to double check on GCP) I'm indifferent to the value used here 😃

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
gene-redpanda: 😅 also same as above

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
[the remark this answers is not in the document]
gene-redpanda: Unfortunately the long weekend will mostly be spent in hospital waiting rooms but I appreciate the sentiment!

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
tmgstevens: Yeah, I don't think it's a big deal either way. What we ought to do is move the client section out into a client role. It's useful for this to work when you're logged into a broker, but it's more useful to have an actual client config that we can deploy to other servers.

gene-redpanda: Not a bad idea. I'll make a note to add that in a later round of work.

I've added some convenience scripting (that could do with a cleanup) to aid in rapid testing and standup that includes some one liners to generate REDPANDA_BROKERS and the like, but it'd make more sense to do something more complete.

ansible/playbooks/roles/redpanda_broker/templates/configs/defaults.j2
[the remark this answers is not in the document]
gene-redpanda: @tmgstevens I've swapped this back to using private IP as it does indeed work with the other changes

tmgstevens:
LGTM

tmgstevens (approved):

PR #137, opened 2023-02-25
adding ci to the repo

gene-redpanda:
I've tested the task stuff pretty thoroughly, including using the test script to validate whether everything can be successfully tested. 

I've yet to test the buildkite pipeline and have low confidence that it is implemented correctly/to spec. 

Zero documentation, and it will need some updating as I've changed how the hosts file is named when generated -- although to provide a stopgap we currently generate two hosts files with one using the original naming schema. The intent is to support multiple simultaneous providers. 

The ca certs are still building in a nasty spot inside the ansible directory and I've yet to clean that up, but it should be a relatively easy fix once I've had a minute to do it and validate that I haven't caused fatal harm to the docs with the change 😅


aws/cluster.tf
```diff
@@ -6,7 +6,7 @@ locals {
   uuid                       = random_uuid.cluster.result
   timestamp                  = time_static.timestamp.unix
   deployment_id              = length(var.deployment_prefix) > 0 ? var.deployment_prefix : "redpanda-${substr(local.uuid, 0, 8)}-${local.timestamp}"
-  tiered_storage_bucket_name = "${local.deployment_id}-bucket"
+  tiered_storage_bucket_name = replace("${local.deployment_id}-bucket", "_", "-")
```
gene-redpanda, line 9: this solved a bug caused by aws not liking underscores

.buildkite/scripts/runtask.sh
```diff
@@ -0,0 +1,11 @@
...
+
+cleanup() {
+  echo "trapped exit, cleaning up"
+  task destroy
+  exit $?
+}
+trap cleanup EXIT INT TERM
+
+task demo-standup
+task destroy
```
ivotron, line 11: fyi, in case it helps, task has a feature for exposing golang's `defer`. works nicely for this type of scenarios https://taskfile.dev/usage/#doing-task-cleanup-with-defer

gene-redpanda:
After a good bit of poking and prodding everything is WAD and WAI! Ready for review 💃 🔮 🕺

.buildkite/scripts/runtask.sh
ivotron: that can still be a single line.  having a high-level `ci` like:

```yaml
tasks:
  ci:
    desc: run CI tasks
    cmds:
      - defer: { task: destroy }
      - demo-setup
      - test-tls-cluster
```

feel free to ignore though... just giving ideas

gene-redpanda: I think in this case I will unfortunately declare stet 😅. Having stranded resources I couldn't clean up myself was a scarring experience and I don't want to risk repeating it -- even if the alternative is cleaner.

ivotron (commented):
nice! would it be possible to get the history rearranged/squashed in logical units of change? that would make it easier to review in chronological/git order. the [contributing page](https://github.com/redpanda-data/redpanda/blob/dev/CONTRIBUTING.md) has some good content and examples.

gene-redpanda:
> nice! would it be possible to get the history rearranged/squashed in logical units of change? that would make it easier to review in chronological/git order. the [contributing page](https://github.com/redpanda-data/redpanda/blob/dev/CONTRIBUTING.md) has some good content and examples.

I've squashed it down and added a more informative commit message!

.buildkite/scripts/runtask.sh
```diff
@@ -0,0 +1,25 @@
...
+trap cleanup EXIT INT TERM
+task ci-standup TF_CLI_ARGS='-var=tags={
+  "vanta-owner" : "devex"
+  "vanta-non-prod" : "true"
+  "vanta-description" : "cicd-instance-for-devex"
+  "vanta-contains-user-data" : "false"
+  "vanta-user-data-stored" : "none"
+  "vanta-no-alert" : "this-is-a-testing-instance-with-no-stored-data"
+}'
+task test-tls-cluster
```
r-vasquez, line 17: So if I wanted to add a new test, this will be the place where I should put the task execution? (It'll be good to add a guide in CONTRIBUTING.md)

Taskfile.yaml
```diff
@@ -0,0 +1,224 @@
...
+    cmds:
+      - |
+        terraform init
+        terraform apply -auto-approve {{.TF_SSH_USER_VAR}} \
+          -var='cloud_provider={{.CLOUD_PROVIDER}}' \
+          -var='deployment_prefix={{.DEPLOYMENT_ID}}' \
+          -var='public_key_path={{.PUBLIC_KEY}}' \
+          -var='nodes={{.NUM_NODES}}' \
+          {{.CLI_ARGS}}
+  #       -backend-config='workspace_dir={{.BUILD_ROOT}}/terraform/state.d/{{.CLOUD_PROVIDER}}/'
```
r-vasquez, line 89: Should we delete this?

.buildkite/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN apt install wget -y
+RUN apt install curl -y
+
+# Install Git
+RUN apt-get install -y git
+
+# Install Ansible
+RUN apt-get install -y software-properties-common && \
+    apt-add-repository --yes --update ppa:ansible/ansible && \
+    apt-get install -y ansible
```
r-vasquez, line 21: [Question] Why use `apt install` in some and `apt-get install` in other layers? Curious.

[Suggestion] Can we merge the RUN layers ?

```
RUN apt-get update && apt-get install -y \
  curl \
  git \
  software-properties-common \
  ...
  // and possibly remove /var/lib/apt/lists/*
```

r-vasquez (commented):
Few non-blocking observations/questions, however it LGTM but I'm not an expert in buildkite so if we can also get a +1 from Devprod would be nice. 

++ Some linter warnings in the Taskfile.yaml

Taskfile.yaml
r-vasquez: Got it, it's worth adding this comment to the code.

.buildkite/Dockerfile
gene-redpanda: I mostly used whatever was in the doc for each vendor. I just checked and merging the layers cut the size by about 10mb on an image that is fairly chonky at 750mb (using ubuntu to match the servers). I'll add it in.

r-vasquez: Nice, it's something :smile:

.buildkite/scripts/runtask.sh
gene-redpanda: I would probably recommend integrating it into the test-tls-cluster task rather than adding a new task (or at least making it a dependency of test-tls-cluster

twmb (approved):

Taskfile.yaml
```diff
@@ -0,0 +1,224 @@
...
+    desc: init and apply terraform
+    summary: |
+      You can set additional arguments to be consumed by terraform for building out the cluster by adding them after 
+      '--'when calling the task. You can set the number of nodes by setting NUM_NODES
+      You may also want to set a backend for your runs if you are using something like Terraform Cloud to manage state. 
+      See the commented out line in cmds for an example of how to achieve this
+    env:
+     AWS_ACCESS_KEY_ID: '{{.DA_AWS_ACCESS_KEY_ID}}'
+     AWS_SECRET_ACCESS_KEY: '{{.DA_AWS_SECRET_ACCESS_KEY}}'
+     AWS_DEFAULT_REGION: '{{ default "us-west-2" .AWS_DEFAULT_REGION}}'
```
ivotron, line 78: note that, due to the way go-task works, environment variables that exist prior to the execution of task cannot be overridden (see https://github.com/go-task/task/discussions/449). when this executes locally, if `AWS_*` exist, they won't be overridden by the values of `DA_*`. it might be worth adding a note or make it part of the `summary:`

Taskfile.yaml
```diff
@@ -0,0 +1,224 @@
...
+      - ansible-galaxy install -r "ansible/requirements.yml"
+
+  basic:
+    desc: run an ansible playbook to create a basic private cluster
+    summary: |
+      You will need to export PRIV_KEY_LOC with the location of the private key matching the public key you provided 
+      in the terraform apply task for the ansible playbook to successfully connect and apply.
+    deps:
+      - install-role-requirements
+    run: always
```
ivotron, line 128: this is only needed when there is a top-level `run: once`

Taskfile.yaml
```diff
@@ -0,0 +1,224 @@
...
+    cmds:
+      - |
+        cat <(echo '[all:vars]
+        tls=true
+        advertise_public_ips=true
+
+        '
+        ) {{.ANSIBLE_INVENTORY}} > tmp.ini
+        rm {{.ANSIBLE_INVENTORY}}
+        mv tmp.ini {{.ANSIBLE_INVENTORY}}
```
ivotron, line 169: the terraform output already generates this, would it be better to generate it via `terraform output` instead?

ivotron (approved):
minor comments, mostly nits

.buildkite/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+RUN apt install unzip -y \
+    && apt install wget -y \
+    && apt install curl -y \
+    && apt-get install -y git
+
+# Install Ansible
+RUN apt-get install -y software-properties-common && \
+    apt-add-repository --yes --update ppa:ansible/ansible && \
+    apt-get install -y ansible
```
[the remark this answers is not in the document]
gene-redpanda: I've modified the dockerfile, total reduction of 60MB on a 750MB image which is something

.buildkite/scripts/installtask.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
gene-redpanda: Yeah it can go

.buildkite/scripts/testcluster.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
gene-redpanda: Just didn't think of doing it that way at the time 🤷‍♀️

Taskfile.yaml
```diff
@@ -0,0 +1,224 @@
...
+      - unzip '{{.TERRAFORM_INSTALL_DIR}}/bin/tf.zip' -d '{{.TERRAFORM_INSTALL_DIR}}/bin'
+      - rm '{{.TERRAFORM_INSTALL_DIR}}/bin/tf.zip'
+      - python3 -mvenv '{{.ANSIBLE_INSTALL_DIR}}'
+      - |
+        source '{{.ANSIBLE_INSTALL_DIR}}/bin/activate'
+        pip3 install --upgrade pip
+        pip3 install ansible-core=={{.ANSIBLE_VERSION}}
+    status:
+      - test -f '{{.TERRAFORM_INSTALL_DIR}}/bin/terraform || command -v terraform'
+      - test -f '{{.ANSIBLE_INSTALL_DIR}}/bin/ansible || command -v ansible'
```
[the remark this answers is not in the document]
gene-redpanda: that task is only there for local install if a user decides to use it for that, it's not part of the ci flow

Taskfile.yaml
gene-redpanda: ok. that's the intended way things should work so I'll note it 👍

Taskfile.yaml
gene-redpanda: So the disadvantage of doing it that way is we can't support BYOB hosts files which I wanted to do on the assumption that some people might use this workflow while not using the terraform side (even if this really only exists for CI and developer testing). 

I may update the terraform output to support different Redpanda vars that ansible can handle in the future so that they flow together better but its a low prio.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,14 @@
+agents:
+  queue: "k8s-builders"
+
+steps:
+  - label: test tls cluster standup
+    key: aws-up
+    command: .buildkite/scripts/runtask.sh
```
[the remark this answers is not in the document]
gene-redpanda: This would honestly be the saner way to do things and if I have a moment tonight I'll get it done. Task calling a script to call task is sorta dumb, but it was my response to fears about builds being canceled -- thereby stranding assets -- and needing a way to guarantee that that doesn't happen.

andrewhsu (approved):
LGTM considering the [build](https://buildkite.com/redpanda/deployment-automation/builds/63) for c50145d was green and it looks like it deployed what it was intending to deploy.

as for cleaning up stranded compute which could be a risk upon every push to a PR, i can see 2 paths:
1. test terraform apply only on manual request like via github label trigger. it would help if the aws resources could have a tag that could point back to the PR and git commit so resource cleanup can be easier.
2. hook up this repo to terraform cloud to deploy only on merge to main and have the PR check lint and terraform plan. then the cloud will keep track of the state file and manage cleanup.

gene-redpanda:
@ivotron thanks for looking things over! BTW is there a way to globally guarantee that only one build of this pipeline can be run at a time? I'd like to keep costs low and having it spam builds because multiple commits came in in quick succession seems dumb. I could randomize the deploy-id and just have multiple going simultaneously but I'd rather they all wait and then the waits be canceled.

ivotron:
> is there a way to globally guarantee that only one build of this pipeline can be run at a time?

that can only be done by creating a new agent pool and setting a limit there . one thing that can be done is to use a [`pre-exit` hook](https://buildkite.com/docs/agent/v3/hooks#job-lifecycle-hooks) that invokes the cleanup routine, and configure a large enough value for the [`cancel-grace-period` agent config option](https://buildkite.com/docs/agent/v3/configuration#cancel-grace-period) so that whenever a build is cancelled, it can properly cleanup resources.

PR #138, opened 2023-02-28
Added optional vpc_id and subnet_id to allow deploying in a specific vpc

genghis-tuan:
Added variables for subnet_id and vpc_id that are defaulted to empty string. An attempt will be made to fetch the default vpc if a vpc_id is not specified. If no subnet_id is specified, it defaults to a subnet in the default vpc.
This allows users to choose a vpc and subnet for the deployment. An error is thrown if the subnet is not in the the vpc.


genghis-tuan:
Closing and recreating. The CLA did not seem to go through.

PR #140, opened 2023-02-28
Added vpc and subnet ID variables and inputs for instances and security group

genghis-tuan:
Added variables for subnet_id and vpc_id that are defaulted to empty string. An attempt will be made to fetch the default vpc if a vpc_id is not specified. If no subnet_id is specified, it defaults to a subnet in the default vpc.
This allows users to choose a vpc and subnet for the deployment. An error is thrown if the subnet is not in the the vpc.


genghis-tuan:
I am not quite sure what to do. I have signed the CLA but it seems to not update. This is my first open source contribution.

genghis-tuan:
closing again because of CLA

PR #149, opened 2023-03-06
Fixed a couple of errors in readme comparing with vars.tf

gfcrawford:
Just made a couple of changes to the readme to reflect what I see in vars.tf.  I think there may be a few additional changes needed in the readme but just want to try the PR process with these first.


gene-redpanda (approved):
Thanks for the awesome contribution!

PR #150, opened 2023-03-07
add consume from topic

gene-redpanda:
without -o :end the topic consumption will hang. hopefully this will cause it to simply dump the last message and exit immediately (if i got it right)


twmb (approved):

gene-redpanda:
did not work as expected

PR #151, opened 2023-03-07
merge main into multi az pr

gene-redpanda:


aws/vars.tf
```diff
@@ -4,6 +4,12 @@ variable "aws_region" {
   default     = "us-west-2"
 }
 
+variable "availability_zone" {
+  description = "The AWS AZ to deploy the infrastructure on"
+  default     = ["us-west-2a", "us-west-2b", "us-west-2c"]
```
gene-redpanda, line 9: it might be nice to default to single az

gcp/vars.tf
```diff
@@ -2,14 +2,15 @@ variable "region" {
   default = "us-west2"
 }
 
-variable "zone" {
+variable "availability_zone" {
   description = "The zone where the cluster will be deployed [a,b,...]"
-  default     = "a"
+  default     = ["us-west2-a", "us-west2-b"]
```
gene-redpanda, line 7: same comment

aws/vars.tf
hcoyote: and add a comment about the cost impact of enabling multi-az

vuldin (changes requested):
I think azure changes should be removed since there are a few issues, and we can have a separate PR for azure afterwards.

azure/network.tf
```diff
@@ -209,7 +209,7 @@ resource "azurerm_public_ip" "redpanda" {
   resource_group_name = azurerm_resource_group.redpanda.name
   location            = azurerm_resource_group.redpanda.location
   allocation_method   = "Static"
-  availability_zone   = try(var.zone, "Zone-Redundant")
+  availability_zone   = try(var.availability_zone[count.index % length(var.availability_zone)], "Zone-Redundant")
```
hcoyote, line 212: Just to make sure I understand what you're doing here:

for everything that has a count based on, like var.client_vm_instances, we're setting the az for that resource instance to whatever az is in the az list based on `Index MOD  az array count`?

So if we have a 3 node cluster, and 3 azs, we get things evenly spread. But if we have a 3 node cluster and 2 azs, we end up with a SPOF on one az because it can contain two brokers. 

We should probably be warning on this somehow.

azure/vars.tf
```diff
@@ -3,9 +3,9 @@ variable "region" {
   default     = "centralus"
 }
 
-variable "zone" {
+variable "availability_zone" {
   description = "Availability Zone"
-  default     = null
+  default     = []
```
hcoyote, line 8: Why default this to no AZs, while we have multi-azs defined in the other clouds?

gcp/cluster.tf
```diff
@@ -61,6 +62,7 @@ resource "google_compute_instance" "monitor" {
   name         = "rp-monitor-${local.deployment_id}"
   tags         = ["rp-cluster", "tf-deployment-${local.deployment_id}"]
   machine_type = var.monitor_machine_type
+  zone         = var.availability_zone[0]
```
hcoyote, line 65: Should this always be pinned to the first listed az? Not sure I understand why we do this here, but allow picking one of the multiple listed in var.availability_zone in other gcp resources

gcp/cluster.tf
```diff
@@ -119,12 +122,36 @@ KEYS
...
+      ],
+      [
+        for j in range(i, length(google_compute_instance.monitor), length(var.availability_zone)) :
+        google_compute_instance.monitor[j].self_link
+      ]
+    )
+
+  }
+}
+
```
hcoyote, line 145: Can you add a comment here to describe what this is doing?

hcoyote (changes requested):

gcp/cluster.tf
gene-redpanda: I'm honestly struggling to understand what this is doing and why it is doing it. I'd have to run it to be sure, but it kinda looks like it is generating self links for each az for each instance even when some of the things (monitor, client) should only be created in one az.

azure/outputs.tf
```diff
@@ -41,6 +41,7 @@ resource "local_file" "hosts_ini" {
       monitor_private_ip         = var.enable_monitoring ? "${azurerm_network_interface.monitoring.0.private_ip_address}" : ""
       enable_monitoring          = "${var.enable_monitoring}"
       ssh_user                   = "${var.admin_username}"
+      availability_zone          = "${azurerm_linux_virtual_machine.redpanda.*.zone}"
```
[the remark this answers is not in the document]
gene-redpanda: Removed all azure content

gcp/cluster.tf
vuldin:
```suggestion
// generate a list of hosts (redpanda instances, clients, and monitor instances) for each availability zone
locals {
  hosts_map = {
    for i in range(length(var.availability_zone)) : var.availability_zone[i] =>
    concat(
      [
        for j in range(i, length(google_compute_instance.redpanda), length(var.availability_zone)) :
        google_compute_instance.redpanda[j].self_link
      ],
      [
        for j in range(i, length(google_compute_instance.client), length(var.availability_zone)) :
        google_compute_instance.client[j].self_link
      ],
      [
        for j in range(i, length(google_compute_instance.monitor), length(var.availability_zone)) :
        google_compute_instance.monitor[j].self_link
      ]
    )

  }
}

```

tmgstevens: Yes, so it's generating a map of AZ-> Host which is then used to generate the instances section of the google_compute_instance_group. Previously we just enumerated all of the brokers, clients and monitoring nodes, but because we could have n instance_groups we now need to be able to iterate across the hosts on a per-AZ basis.

azure/network.tf
tmgstevens: Your commentary is correct. `Index mod AZ Array Count` is how we're doing things. I'm not sure we need to comment this specifically - hopefully it's obvious that if you put 3 things in 2 places then they aren't evenly split

gcp/cluster.tf
tmgstevens: Added a comment explaining this and why we do it

tmgstevens:
Fixed by #165

PR #156, opened 2023-03-16
Push tasks into the role

gene-redpanda:
The intent of this PR is to push most of the "assisting" playbooks into the role to provide a single stop shop for building out a cluster. The end result is playbooks that clearly articulate a single final objective and have everything necessary to get it done, simplifying the experience of using the repo.

Note that it is still entirely possible to work flexibly by importing different task files from the role to achieve different goals (as is demoed with the TLS cluster).

The default path for the role is still the simplest possible option: installing and starting the RP node. In future it may be prudent to expand the default, but for now this seems correct.

Impact on current users:

Should be fairly minimal although it will definitely break any shell scripts. Tasks are still done the same way, they're just called differently. I'll provide some additional mitigation by tagging our current main as v1.0.0, with this edition tagged as v2.0.0, clearly signaling a potentially breaking change.

edit: pulling from most recent commit

Starting with the simple stuff, I moved all the roles out of playbooks and placed them directly under ansible to improve clarity since we no longer have so many playbooks.

Next I moved the invocation of the data dir and dependencies installs into main.yaml of redpanda_broker. I felt this was a more sensible location, especially as we can control whether or not they fire by using the trigger variables.

I also moved all cert related content into the demo_certs role. I wanted the name to clearly indicate that this was NOT a production use role. There are no functional changes to the content in this role.

Finally I updated the provision-tls-cluster role to reflect the new changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,9 +117,9 @@ ansible-playbook ansible/playbooks/provision-node.yml -i hosts.ini --extra-vars 
 
 * The Grafana URL is http://&lt;grafana host&gt;:3000/login
 
-## Configure Graphana
+## Configure Grafana
 
-To deploy a graphana node, ensure that you have a [monitor] section in your hosts file. You should then be able to run
+To deploy a grafana node, ensure that you have a [monitor] section in your hosts file. You should then be able to run
 the deploy-prometheus-grafana.yml playbook
 
 ```shell
@@ -143,10 +143,10 @@ For example:
 ansible-playbook ansible/playbooks/provision-tls-cluster.yml \
 -i hosts.ini \
 --private-key '<path to a private key with ssh access to the hosts>' \
---extra-vars redpanda_key_file='<path to key file>' \
---extra-vars redpanda_cert_file='<path to cert file>' \
---extra-vars redpanda_truststore_file='<path to truststore file> \
---extra-vars handle_certs=false'
+--extra-vars create_demo_certs=false \
+--extra-vars advertise_public_ips=false \ 
+--extra-vars handle_certs=false \
+--extra-vars redpanda_truststore_file='<path to ca.crt file>' 
 ```
 
 The second option is to deploy a local certificate authority using the playbooks provided below and generating private
@@ -162,24 +162,8 @@ ansible-playbook ansible/playbooks/provision-tls-cluster.yml \
 
 ## Adding Nodes to an existing cluster
 
-The playbooks can be used to add nodes to an existing cluster however care is required to make sure that they playbooks
-are executed in the correct order. To add new nodes execute the playbooks in the following order:
-
-1. Add the new host(s) to the `hosts.ini` file. You may add `skip_node=true` to the existing hosts to avoid the
-   playbooks being re-run on those nodes.
-2. `install-node-deps.yml` - this will set up the Prometheus node_exporter and install package dependencies.
-3. `prepare-data-dir.yml` - this will create any RAID devices required and format devices as XFS. Note: This playbook
-   looks for devices presented to the operating system as NVMe devices (which can include EBS volumes built on the Nitro
-   System). You may replace this playbook with your own method of formatting devices and presenting disks.
-4. If managing TLS with the Redpanda playbooks:
-1. `generate-csrs.yml` - will create private key and CSR and bring the CSR back to the Ansible host.
-2. If using the Redpanda provided CA: `issue-certs.yml` - signs the CSR and issues a certificate.
-3. `install-certs.yml` - Installs the certificate and also applies the `redpanda_broker` role to the cluster nodes.
-   Note: This will install and start Redpanda (and restart any brokers that do not have `skip_node=true` set)
-5. If `install-certs.yml` was not run in step iii above, you will need to run `provision-node.yml` which will install
-   the `redpanda_broker` role onto any nodes without `skip_node=true` set. **Note: If TLS is enabled on the cluster,
-   make sure that `-e tls=true` is set, otherwise this playbook will disable TLS across any nodes that don't
-   have `skip_nodes=true` set.**
+To add nodes to a cluster you must add them to the hosts file and run the relevant playbook again. You may
+add `skip_node=true` to the existing hosts to avoid the playbooks being re-run on existing nodes.
 
 ## Upgrading a cluster
 
@@ -191,7 +175,7 @@ cluster.
 
 > Note: Please be aware that any changes that have been made to cluster or node configuration parameters outside of the
 > playbook may be overwritten by this procedure, and therefore these settings should be incorporated as part of the
-> provided `--extra-vars` (for example `--extra-vars=tls=true`).
+> provided `--extra-vars` (for example `--extra-vars=enable_tls=true`).
 
 There are two ways of enacting an upgrade on a cluster:
 


gene-redpanda:
https://buildkite.com/redpanda/deployment-automation/builds/90

gene-redpanda:
Building clean, will deal with linter later 🙄  https://buildkite.com/redpanda/deployment-automation/builds/91#01870595-ecbc-41a4-9cc4-3be41b2e6c74

Taskfile.yaml
```diff
@@ -221,7 +181,17 @@ tasks:
...
     vars:
       RPK_LOC: '{{.ARTIFACT_DIR}}/bin/rpk'
-      CA_CRT: '{{ default "ansible/playbooks/tls/ca/ca.crt" .CA_CRT }}'
+      CA_CRT: '{{ default "ansible/tls/ca/ca.crt" .CA_CRT }}'
+      ANSIBLE_INVENTORY: '{{.ARTIFACT_DIR}}/hosts_{{.CLOUD_PROVIDER}}_{{.DEPLOYMENT_ID}}.ini'
+    cmds:
+      - '{{.TASKFILE_DIR}}/.buildkite/scripts/test_tls_cluster.sh --hosts={{.ANSIBLE_INVENTORY}} --cert={{.CA_CRT}} --rpk={{.RPK_LOC}}'
+
+  test-basic-cluster:
+    desc: tests that a public ip TLS cluster is WAD
```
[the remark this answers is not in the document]
r-vasquez: Working as desired?

ansible/roles/demo-certs/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
+# creates a local CA for demo purposes, should not be used in prod
```
r-vasquez, line 1: [question] if `demo_cert` is the **role** needed to handle certs (local CA or ext provided) why is it called `demo_*`?

I think I understand that we also include the local CA creation but it also includes the configuration settings (inside **install-certs**) which will be needed for production users.

.buildkite/scripts/test_basic_cluster.sh
```diff
@@ -0,0 +1,51 @@
...
+"${PATH_TO_RPK_FILE}" cluster status --brokers "$REDPANDA_BROKERS" -v || exit 1
+
+echo "creating topic"
+"${PATH_TO_RPK_FILE}" topic create testtopic --brokers "$REDPANDA_BROKERS" -v || exit 1
+
+echo "producing to topic"
+echo test | "${PATH_TO_RPK_FILE}" topic produce testtopic --brokers "$REDPANDA_BROKERS" -v || exit 1
+
+echo "consuming from topic"
+"${PATH_TO_RPK_FILE}" topic consume testtopic --brokers "$REDPANDA_BROKERS" -v -o :end || exit 1
```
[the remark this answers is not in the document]
gene-redpanda: oh thanks for reminding me to delete those, i completely forgot

Taskfile.yaml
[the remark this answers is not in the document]
gene-redpanda: working as designed, which often implies not working as desired 😂

ansible/roles/demo-certs/tasks/install-certs.yml
```diff
@@ -0,0 +1,32 @@
...
+  tags:
+    - install_certs
+  set_fact:
+    tls: true
+
+- name: install_certs - Apply updated config via reapplying the redpanda_broker role
+  tags: install_certs
+  import_role:
+    name: redpanda_broker
+  when: not skip_node | default(false) | bool
```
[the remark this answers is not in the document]
gene-redpanda: There's a lot of nasty stuff that happens with dependencies in playbooks so I'd prefer to avoid explicit ones unless there's a really serious advantage. More info on some imports that caused bugs [in this commit ](https://github.com/redpanda-data/deployment-automation/pull/159/commits/520208224ab09aa939394d799fd4bf60a77b2093)on the linter PR

ansible/roles/demo-certs/tasks/main.yml
gene-redpanda: I want to strongly discourage using this module in production, so I've given it a name that will cause people to really strongly consider whether they want to use it vs simply handling their own certs. 

The install-certs import is actually incorrect and I've removed it in the other PR -- I know it's confusing, but I didn't want to clobber this one with linter changes.

.buildkite/scripts/test_basic_cluster.sh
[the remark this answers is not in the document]
gene-redpanda: done

r-vasquez (approved):
LGTM, Linter stuff will be in other PR :+1:

ansible/provision-node.yml
```diff
@@ -0,0 +1,15 @@
+# This playbook provisions a Redpanda cluster with no bells or whistles for evaluation purposes
+# variables
+# advertise_public_ips : causes the cluster to set a public ip on advertised_kafka_api, allowing access to the cluster from the public internet
+---
+- name: provision nodes
+  hosts: redpanda
+  vars:
+    advertise_public_ips: true
+    prep_data_dir: true
+    prep_deps: true
```
hcoyote, line 10: I think we should be setting the default to true for the node prep stuff in it's own tasks rather than doing it this way. I can't think of a reason where we would really ever turn this off.

ansible/roles/redpanda_broker/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
+# installs necessary dependencies for running redpanda
+- name: Install dependencies
+  ansible.builtin.include_tasks: install-node-deps.yml
+  when: prep_deps | default(false) | bool
```
hcoyote, line 4: This should be true by default. Users should opt-out of this rather than having to opt-in.

ansible/roles/redpanda_broker/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+
+# configures the data dir. won't work if Install dependencies isn't run first
+- name: Prep data dir
+  ansible.builtin.include_tasks: prepare-data-dir.yml
+  when:
+    - prep_data_dir | default(false) | bool
```
hcoyote, line 10: This should be true by default. Users should opt-out of this rather than having to opt-in.

ansible/roles/redpanda_broker/tasks/prepare-data-dir.yml
```diff
@@ -0,0 +1,86 @@
...
+    - name: Create XFS file system
+      community.general.filesystem:
+        fstype: xfs
+        dev: "{{ nvme_devices_for_raid[0] }}"
+    - name: Mount NVMe device
+      ansible.builtin.mount:
+        path: /mnt/vectorized
+        src: "{{ nvme_devices_for_raid[0] }}"
+        fstype: xfs
+        opts: "{{ 'defaults,nofail' if ephemeral_disk else 'defaults'}}"
```
hcoyote, line 60: @tmgstevens should this include noatime/nodiratime?

ansible/roles/redpanda_broker/tasks/prepare-data-dir.yml
```diff
@@ -0,0 +1,86 @@
...
+        state: mounted
+      when: nvme_devices_for_raid|length == 1
+
+- name: Create data path
+  ansible.builtin.file:
+    path: /mnt/vectorized/redpanda
+    state: directory
+    owner: root
+    group: root
+    mode: ugo=rwx
```
hcoyote, line 72: Doesn't this result in permissions that are mode 777?

ansible/roles/redpanda_broker/tasks/prepare-data-dir.yml
```diff
@@ -0,0 +1,86 @@
...
+
+
+- name: Set permissions
+  ansible.builtin.file:
+    dest: /var/lib/redpanda
+    src: /mnt/vectorized/redpanda
+    state: link
+    owner: root
+    group: root
+    mode: ugo=rwx
```
hcoyote, line 84: same mode 777 comment as above.

ansible/roles/redpanda_broker/tasks/install-node-deps.yml
```diff
@@ -0,0 +1,54 @@
...
+  tags:
+    - node_deps
+  block:
+    - name: delete unused repos
+      ansible.builtin.file:
+        name: "{{item}}"
+        state: absent
+      with_items:
+        - /etc/yum.repos.d/fedora-updates-modular.repo
+        - /etc/yum.repos.d/fedora-cisco-openh264.repo
```
hcoyote, line 19: I know these were here before. but I don't know why we're doing this.

hcoyote (dismissed):

ansible/provision-node.yml
gene-redpanda: It isn't included by default in the original role, so I didn't want to include it by default in the new role. I also think there is a clear scenario where someone would want these off: golden image with all necessary RP deps and R2G data dir. If someone had that and we started mucking about, they'd be annoyed.

ansible/roles/redpanda_broker/tasks/prepare-data-dir.yml
gene-redpanda: That's what it was set to do I'm loathe to change it even if I agree it isn't great

gene-redpanda: See the linter PR for explicit mode setting

ansible/roles/redpanda_broker/tasks/install-node-deps.yml
gene-redpanda: Cargo cult rn, I haven't had a chance to test without some of this stuff.

tmgstevens:
I'd still like to review this. I'll try to get to it tomorrow if that's okay?

gene-redpanda:
> I'd still like to review this. I'll try to get to it tomorrow if that's okay?

No worries,I made sure to reserve this week for reviews and won't be merging til Monday morning at the earliest (after correcting for requested changes).

ansible/roles/redpanda_broker/tasks/prepare-data-dir.yml
hcoyote: if this is generating 777, this is a security hole.

ansible/roles/redpanda_broker/tasks/install-node-deps.yml
hcoyote: No problem. let's flag it for review later. this shouldn't holdup the pr.

ansible/provision-node.yml
hcoyote: I think this is one area where we need to be opinionated for supportability reasons. We don't want people creating mounts that aren't xfs, for example.

ansible/roles/redpanda_broker/tasks/prepare-data-dir.yml
gene-redpanda: I don't have a good way to test the ramifications of a change here barring standard "can we produce and consume a message from the cluster". If you can recommend one I'd be much happier making a change. 

I don't think it will have an impact changing it to 644 or something, but I'm not able to verify it.

ansible/provision-node.yml
gene-redpanda: I think if we were starting greenfield I'd agree, but we aren't. So it doesn't make sense to have functionality enabled that wasn't enabled in the past.

ansible/roles/redpanda_broker/tasks/prepare-data-dir.yml
hcoyote: it's a directory, so it'll need to be 755 at least.

ansible/roles/redpanda_broker/tasks/prepare-data-dir.yml
hcoyote: realized this one is a link on a second look. the mode doesn't really matter here, it'll always be 777 on the filesystem, so think we can ignore this one.

ansible/roles/redpanda_broker/tasks/prepare-data-dir.yml
gene-redpanda: I guess another way to put it is that there were only three file/directory creation tasks that have a mode set and this 777 is two of them. I don't know enough yet to say whether that's because someone threw their hands up and yolo'd or because it is actually necessary. I don't think it'd be a problem changing it, but I don't know what edge cases I haven't thought of yet 😓

hcoyote: so there's going to me two that I'm aware of:

the directory created where we overlay the filesystem for RP. This should definitely be 755 (and arguably, it should be immutable to prevent certain failure conditions from causing rootfs to fill -- but immutable is not important right now).

the second is the top level directory on the filesystem we create. This should be 755 as well since only redpanda shouldbe writing to it. Challenge is from an ordering perspective. you have to ensure the mount point directory gets created and chmod'd, then mount the fs, then re-do the operation. In practice, it sort of happens on subsequent runs because of the mount overlay.

gene-redpanda: changed to 775 in this PR https://github.com/redpanda-data/deployment-automation/pull/159/commits/a0f508df18d8dc1bd576772dd2d70f303ed57ea3

ansible/provision-node.yml
gene-redpanda: resolved in this commit https://github.com/redpanda-data/deployment-automation/pull/156/commits/a0e885c35a43be527c844671a05dde3a9bafcbf8

.pre-commit-config.yaml
```diff
@@ -1,24 +1,24 @@
...
+      - id: trailing-whitespace
+        files: '\.(yml|yaml)$'
+  - repo: https://github.com/ansible/ansible-lint
+    rev: v6.13.1
+    hooks:
+      - id: ansible-lint
+        args: [ "-c", ".ansible-lint.yml" ]
+        files: '\.(yml|yaml)$'
+        language_version: python3
+  #        exclude: ansible/prepare-data-dir.yml
```
twmb, line 18: Should this dead code be deleted?

Dockerfile
```diff
@@ -24,5 +25,9 @@ RUN wget https://releases.hashicorp.com/terraform/1.3.7/terraform_1.3.7_linux_am
 # Install task
 RUN curl -sSLf "https://github.com/go-task/task/releases/download/v3.21.0/task_linux_amd64.tar.gz" | tar -xz -C /usr/local/bin
 
+## uncomment for use as a local client
```
twmb, line 28: Can this comment be expanded a bit? not sure what it means offhand when reading it -- also the following block is already not commented?

README.md
```diff
@@ -38,10 +43,20 @@ To use existing infrastructure, update the `hosts.ini` file with the appropriate
...
+> rolling restart of the cluster to occur. If you do not want the cluster to restart you can
+> specify `restart_node=false`
+> either globally (as --extra-vars, or on a host-by-host basis in the `hosts.ini` file). If you make changes to the node
+> config and do not perform a restart then you may leave your `rpk` config (and this may inhibit subsequent executions
+> of
+> the playbook). If you re-run the playbook it will overwrite any configurations with those specified in the playbook,
+> so
+> care should be taken to ensure that the playbook contains any desired configuration (for example, if you have enabled
+> TLS on your cluster, any subsequent runs of `provision-node` should be made with `-e tls=true` otherwise the playbook
+> will disable TLS).
```
twmb, line 56: +1 to the wrapping, however lines 51 and 53 now seem oddly short?

ansible/provision-tls-cluster.yml
```diff
@@ -0,0 +1,29 @@
+# Creates a TLS enabled cluster with local CA in artifacts/ca for testing purposes
+# !!! WARNING !!!
+# If you intend to use this for a production cluster you will need to set handle_certs to false and provide your own certs
+# Using our demo CA for production is a bad idea :S
```
twmb, line 4: change to
```
Our demo CA should not be used for production or your cluster will be insecure!
```

Dockerfile
gene-redpanda: I use the same image for the testing machines and for local connection to the testing machines as it is more convenient than clearing out my macbook every time. I forgot to stash the commenting out this go round

.pre-commit-config.yaml
gene-redpanda: I'm still holding out hope getting the ansible-lint pre-commit working 😞 and that commented line is necessary for it to not explode

Dockerfile
gene-redpanda: Fixed!

ansible/provision-tls-cluster.yml
hcoyote: Sorry, why is this insecure? We shouldn't use language like that.

gene-redpanda: In fairness it's relatively true. A self signed cert from a CA on your local laptop is super bad practice and not what I think anyone would consider secure. Plus we aren't exactly doing much to set the local CA up for success

hcoyote: There's nothing insecure about it. It creates properly formatted TLS certificates for use with fully encrypted connections.

The bad part of this is that you have to manage client CA integration manually. If you want to focus on that as the challenge for production environments, no problem. But it's not insecure.

twmb: Does the generated CA have SANS properly setup?

If so, then I can see the argument for this being secure enough -- it's common enough to generate a CA & the certs and then put the CA into a vault.

hcoyote: Let's just say "Using our demo certificate authority for production is not recommended, and is intended only for testing and proof of concepts. Redpanda encourages you to integrate with your corporate certificate management setup."

gene-redpanda: done

tmgstevens (dismissed):
Okay, I'm happy with this. Please can we make sure the README is up to date and works correctly. Once that's updated let's get this merged.

README.md
```diff
@@ -93,98 +112,116 @@ ansible-playbook ansible/playbooks/provision-node.yml -i hosts.ini --extra-vars
...
+NOTE THAT THIS SHOULD ONLY BE DONE FOR TESTING PURPOSES! Use an actual signed cert from a valid CA for production!
 
-`ansible-playbook --private-key <your_private_key> -i hosts.ini -v ansible/playbooks/install-certs.yml`
+```shell
+ansible-playbook ansible/playbooks/provision-tls-cluster.yml \
+-i hosts.ini \
+--private-key '<path to a private key with ssh access to the hosts>'
+```
 
 ## Adding Nodes to an existing cluster
```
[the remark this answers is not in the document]
gene-redpanda: provision-tls-cluster is new, but provision-node still exists. I had thought about renaming it but it still does exactly the same job as it did before so there was no real reason.

twmb (approved):

tmgstevens (approved):

PR #158, opened 2023-03-22
Update tiered_storage.j2 to use correct cloud_storage_region

vuldin:
This seems to be a problem introduced when handling a merge conflict (or possibly with the tiered storage PR itself). The variable was still set as the older (now renamed) `aws_region`.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204218351695001


tmgstevens:
@vuldin can you check that this shouldn't be `cloud_storage_region` (which is what I'm seeing in the hosts_ini.tpl

vuldin:
I had named the variable correctly in the commit message but not the commit :D

gene-redpanda (approved):

PR #171, opened 2023-04-25
Minor additions to AWS TF

gene-redpanda:
Associate public IP fixes an issue some customers had with using a custom VPC while also wanting to use public IPs.

allow_force_destroy allows destroying a bucket even with data in it. Useful for testing and demos.

Everything else is minor text/syntax cleanup.

### DIFF
--- a/aws/cluster.tf
+++ b/aws/cluster.tf
@@ -76,17 +76,19 @@ resource "aws_instance" "redpanda" {
   instance_type              = var.instance_type
   key_name                   = aws_key_pair.ssh.key_name
   iam_instance_profile       = var.tiered_storage_enabled ? aws_iam_instance_profile.redpanda[0].name : null
-  vpc_security_group_ids     = concat([aws_security_group.node_sec_group.id], var.security_groups_redpanda)
   placement_group            = var.ha ? aws_placement_group.redpanda-pg[0].id : null
   placement_partition_number = var.ha ? (count.index % aws_placement_group.redpanda-pg[0].partition_count) + 1 : null
   availability_zone          = var.availability_zone[count.index % length(var.availability_zone)]
-  subnet_id                  = var.subnet_id
   tags = merge(
     local.merged_tags,
     {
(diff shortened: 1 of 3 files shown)


twmb (approved):

aws/s3.tf
```diff
@@ -2,6 +2,7 @@ resource "aws_s3_bucket" "tiered_storage" {
   count  = var.tiered_storage_enabled ? 1 : 0
   bucket = local.tiered_storage_bucket_name
   tags   = local.instance_tags
+  force_destroy = var.allow_force_destroy
```
tmgstevens, line 5: Does this lint okay?

aws/cluster.tf
```diff
@@ -76,18 +76,20 @@ resource "aws_instance" "redpanda" {
...
   availability_zone          = var.availability_zone[count.index % length(var.availability_zone)]
-  subnet_id                  = var.subnet_id
   tags = merge(
     local.merged_tags,
     {
       Name = "${local.deployment_id}-node-${count.index}",
     }
   )
 
+  vpc_security_group_ids      = concat([aws_security_group.node_sec_group.id], var.security_groups_redpanda)
```
tmgstevens, line 89: Slightly confused as to why this moved. As far as I can see there's no change?

aws/cluster.tf
```diff
@@ -76,18 +76,20 @@ resource "aws_instance" "redpanda" {
...
+  subnet_id                   = var.subnet_id
+  associate_public_ip_address = var.associate_public_ip_addr
```
tmgstevens, line 91: Should these attributes be up with the rest of the non-structured elements?

tmgstevens (approved):
I've not tested any of this but it looks okay. Grateful if you could take a quick look at the syntactic stuff

aws/cluster.tf
gene-redpanda: done

aws/cluster.tf
gene-redpanda: moved back

PR #172, opened 2023-04-27
Put needed vars in playbook

gene-redpanda:
Since we're removing these as defaults from the role they should be provided in the playbook. It might be sensible to set version to blank to ensure the user MUST select it.

related to https://github.com/redpanda-data/redpanda-ansible-collection/pull/8

### DIFF
--- a/ansible/provision-node.yml
+++ b/ansible/provision-node.yml
@@ -6,6 +6,7 @@
   hosts: redpanda
   vars:
     advertise_public_ips: true
+    redpanda_version: latest
   tasks:
     - name: Install and start redpanda
       ansible.builtin.include_role:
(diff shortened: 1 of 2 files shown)


twmb:
Is there downside to defaulting to latest? If I leave this latest and re-apply my playbook after a new RP release, does that cause an upgrade?

gene-redpanda:
> Is there downside to defaulting to latest? If I leave this latest and re-apply my playbook after a new RP release, does that cause an upgrade?

My understanding is yes, because the version has changed.

twmb (approved):

vuldin (approved):
LGTM

vuldin:
Actually, should this also be added to the README?

vuldin (commented):
Should this be added to the README?

PR #174, opened 2023-05-10
add missing resourced needed for ListBucket

hcoyote:
the policy we create on behalf of the user for allowing the instance to access the s3 bucket appears to be missing a resource required by Redpanda needed for ListBucket. 

See internal discussion at https://redpandadata.slack.com/archives/C02BDN76HUK/p1683670439226129


gene-redpanda (approved):
LGT

PR #178, opened 2023-05-26
Adding private network capability to Azure.

pmw-rp:
Added flags to allow public ip address creation to be suppressed, along with CIDR customisation.


gene-redpanda (approved):
LGTM! Thanks for contribution!

PR #185, opened 2023-06-01
getting module into action

gene-redpanda:
Moving us from local TF files to an AWS terraform module for managing the cluster.


twmb:
This is a big change -- any end user change?

gene-redpanda:
@twmb so yes and no. the big end user change will be that there is a module and they might need to update their variables some. It may also require some state imports for people using the current setup which is potentially destructive if done really wrong but otherwise a fairly simple process of 

```terraform mv blah module.redpanda-cluster.blah```

where blah is the prior resource's name. Here's an example from some blog: https://ryaneschinger.com/blog/terraform-state-move/

I intend to mark this as 3.0 to indicate that it does have potentially destructive consequences. 

The advantage is that in future users can manage their upgrade with module version numbers and can easily roll back/forward/pin without any fiddling around.

twmb (approved):

twmb:
^^Please create a "How to migrate from v2 to v3" checklist though, as follow up

gene-redpanda:
Fair shout, migration guide is sensible

gene-redpanda:
> ^^Please create a "How to migrate from v2 to v3" checklist though, as follow up

This is done

r-vasquez (approved):

PR #188, opened 2023-06-13
Fix public key

r-vasquez:
This PR contains:

- A quick fix for using the public key as default in TF aws variables.
- (proposal) A basic usage section in the Readme.


gene-redpanda (approved):
THANKS!!!!! This is awesome!

PR #192, opened 2023-06-22
fix for aws standard main.tf

gene-redpanda:
made it easier for users to pass in information necessary to start up a cluster with TF

standardized the inventory file declaration in taskfile to a single variable because why

### DIFF
--- a/aws/main.tf
+++ b/aws/main.tf
@@ -1,17 +1,31 @@
+## we assume a default vpc. if you have one you want to use you will need to provide a vpc and subnet ID
+
 module "redpanda-cluster" {
-  source                 = "redpanda-data/redpanda-cluster/aws"
-  version                = "~> 0.1"
-  public_key_path        = var.public_key_path
-  nodes                  = var.nodes
-  deployment_prefix      = var.deployment_prefix
-  enable_monitoring      = var.enable_monitoring
-  tiered_storage_enabled = var.tiered_storage_enabled
-  allow_force_destroy    = var.allow_force_destroy
-  vpc_id                 = var.vpc_id
-  distro                 = var.distro
-  hosts_file             = var.hosts_file
-  tags                   = var.tags
-  availability_zone      = var.availability_zone
+  source                   = "redpanda-data/redpanda-cluster/aws"
+  version                  = "~> 0.1"
+  public_key_path          = var.public_key_path
+  nodes                    = var.nodes
+  deployment_prefix        = var.deployment_prefix
+  enable_monitoring        = var.enable_monitoring
+  tiered_storage_enabled   = var.tiered_storage_enabled
+  allow_force_destroy      = var.allow_force_destroy
+  vpc_id                   = var.vpc_id
+  distro                   = var.distro
+  hosts_file               = var.hosts_file
+  tags                     = var.tags
+  aws_region               = var.aws_region
+  associate_public_ip_addr = var.associate_public_ip_addr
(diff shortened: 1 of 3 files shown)


r-vasquez:
What was the original issue that made you create this fix PR ?

aws/main.tf
```diff
@@ -87,17 +96,16 @@ terraform {
...
-variable "region" {
+variable "subnet_id" {
+  default = ""
   type    = string
-  default = "us-west-2"
 }
 
-provider "aws" {
-  region = var.region
+variable "aws_region" {
```
r-vasquez, line 104: Do we need to change the naming of the variable here? This might break users and also I think `region` it's fine given that we are running the terraform script in aws.
 
 In the past, there was some work done trying to standardize TF variable naming across different cloud providers and I think that's why we had only `region` here.

aws/main.tf
```diff
@@ -87,17 +96,16 @@ terraform {
...
+  default = "us-west-2"
 }
 
-variable "availability_zone" {
-  description = "The AWS AZs to deploy the infrastructure on"
-  default     = ["us-west-2a"]
-  type        = list(string)
+provider "aws" {
+  region = var.aws_region
 }
```
r-vasquez, line 111: Why delete the availability_zone variable?

gene-redpanda:
@r-vasquez did a call with Denis yesterday where he tried to use this stuff and he ran into a few issues caused by not having a few values/some confusion with docs. I also noticed that I was duplicating the declaration of the hosts file location multiple times so I figured I'd clean that up too.

deniscoady:
We worked through some issues and realized that we probably need to add a few new variables by default:
* We allow people to set `vpc_id`, but they also would be required to set a `subnet_id` (which doesn't have a preset variable).
* We do not pass the aws_region to aws so we needed to add in the module definition: `aws_region = var.region`
* By default none of the boxes were assigned public IPs even if the subnet was told to assign by default. We had to add `associate_public_ip_addr = true` to the module definition.

aws/main.tf
```diff
@@ -1,22 +1,31 @@
...
+
+variable "associate_public_ip_addr" {
+  default = true
+  type    = bool
 }
 
 variable "public_key_path" {
   type    = string
-  default = "~/.ssh/id_rsa.pub"
+  default = "../artifacts/testkey.pub"
```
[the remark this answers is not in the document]
gene-redpanda: fixed, unintended change

aws/main.tf
gene-redpanda: fixed, not sure how that got removed

aws/main.tf
gene-redpanda: I'm fine with it either way, but I worry it could be confusing with the module referring to aws_region while this refers to region.

r-vasquez: I'll leave it up to you, I just don't like making breaking changes as you may have noticed :rofl: 

I'll approve

r-vasquez (approved):

aws/main.tf
gene-redpanda: My viewpoint is that breaking stuff en route to consistency is worth it, especially when 

* these are examples rather than something people depend on
* rolling back or making a change to accommodate it is simple

Both are true in this case.

twmb: So long as we have proper version bumping.

twmb (approved):

PR #196, opened 2023-07-21
adds support for airgap and proxy

gene-redpanda:


ansible/provision-private-proxied-cluster-airgap-rpm.yml
```diff
@@ -0,0 +1,57 @@
...
+    rpm_or_deb: "rpm"
+    ansible_ssh_common_args: "-o ProxyCommand='ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /app/artifacts/testkey -W %h:%p -q {{ hostvars[groups['client'][0]].ansible_user }}@{{ hostvars[groups['client'][0]].inventory_hostname }}'"
+  tasks:
+    - name: Install and configure proxy on client
+      ansible.builtin.include_role:
+        name: redpanda.cluster.binary_bundler
+      run_once: true
+
+## clients[0] is obviously pretty ugly, need to come up with an az friendly way to handle this
+## BUT because it is in the playbook it is easy for the user to adjust to their liking at least
```
r-vasquez, line 30: Can you change the wording here, please? maybe add a TODO with the specific linked issue for future ref.

ansible/provision-private-proxied-cluster.yml
```diff
@@ -0,0 +1,50 @@
...
+  hosts: client
+  vars:
+    squid_acl_localnet:
+      - "10.0.1.0/24"
+  tasks:
+    - name: Install and configure proxy on client
+      ansible.builtin.include_role:
+        name: mrlesmithjr.squid
+
+## clients[0] is obviously pretty ugly, need to come up with an az friendly way to handle this
```
r-vasquez, line 23: ditto

ansible/provision-private-proxied-cluster-airgap-deb.yml
```diff
@@ -0,0 +1,57 @@
...
+    redpanda_version: "23.1.13-1" # note that the -1 is the build number for this version. it is typically -1 but if the version is incremented it will be higher
+    rpm_or_deb: "deb"
+    ansible_ssh_common_args: "-o ProxyCommand='ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /app/artifacts/testkey -W %h:%p -q {{ hostvars[groups['client'][0]].ansible_user }}@{{ hostvars[groups['client'][0]].inventory_hostname }}'"
+  tasks:
+    - name: Install and configure proxy on client
+      ansible.builtin.include_role:
+        name: redpanda.cluster.binary_bundler
+      run_once: true
+
+## clients[0] is obviously pretty ugly, need to come up with an az friendly way to handle this
```
r-vasquez, line 29: ditto

ansible/provision-private-proxied-cluster-airgap-rpm.yml
```diff
@@ -0,0 +1,57 @@
+### This is not a truely airgapped cluster!!! The intent here is specifically to test the install flow for an airgapped install
```
r-vasquez, line 1: It's unclear whether this playbook is a test, an example, or something our clients can use. If it's an example, can we move it to another directory? `examples/` perhaps.

aws/standard/main.tf
```diff
@@ -1,10 +1,9 @@
 ## we assume a default vpc. if you have one you want to use you will need to provide a vpc and subnet ID
 
 module "redpanda-cluster" {
-  source                   = "redpanda-data/redpanda-cluster/aws"
-  version                  = "~> 0.1"
+  source                   = "git::github.com/redpanda-data/terraform-aws-redpanda-cluster.git?ref=private-ip-and-zones"
```
r-vasquez: Will `aws/standard` be our new default way to create a redpanda cluster? if that's correct, why reference `private-ip-and-zones` ?

Also, please update the readme so we point our users to the correct aws folder for the quickstart.

r-vasquez (approved):

aws/standard/main.tf
r-vasquez: Oh, you updated this in another commit later. Still, please update the readme with the new path :+1:

gene-redpanda: This is a testing value that will revert to the standard TF module once I've cut a release (which will happen very soon).

ansible/provision-private-proxied-cluster-airgap-rpm.yml
gene-redpanda: Not sure what happened but my comments I made by phone last week all appear to have been gobbled 🤦‍♀️ anyway, the intent here is to satisfy two different needs:
* the need for a testable playbook for airgap features
* the need for an example for users to show them how we can use the airgap features

I've included the comment to indicate that this isn't actually an airgapped cluster, but I don't have a good idea how to communicate the rest of that just yet.

ansible/bundle-deb-for-airgap.yml
```diff
@@ -0,0 +1,14 @@
...
+- name: bundle debs
+  hosts: client
+  vars:
+    basearch: "amd64" # must match the output of `uname -m`. however on ubuntu you need to use the debian version of the output. So for example x86_64 is amd64
+    os_distribution_major_version: "focal" # must match the output of `. /etc/os-release && echo ${VERSION_ID%%.*}`
+    os_distribution: "ubuntu" # must match the output of `. /etc/os-release && echo $ID`
+    redpanda_version: "23.1.13-1" # note that the -1 is the build number for this version. it is typically -1 but if the version is incremented it will be higher
+    rpm_or_deb: "deb"
+  tasks:
+    - name: Install and configure proxy on client
```
hcoyote, line 11: name seems at odds with playbook.

ansible/bundle-rpm-for-airgap.yml
```diff
@@ -0,0 +1,14 @@
...
+- name: bundle rpms
+  hosts: client
+  vars:
+    basearch: "x86_64" # must match the output of `uname -m`
+    os_distribution_major_version: "34" # must match the output of `. /etc/os-release && echo ${VERSION_ID%%.*}`
+    os_distribution: "fedora" # must match the output of `. /etc/os-release && echo $ID`
+    redpanda_version: "23.1.13-1" # note that the -1 is the build number for this version. it is typically -1 but if the version is incremented it will be higher
+    rpm_or_deb: "rpm"
+  tasks:
+    - name: Install and configure proxy on client
```
hcoyote, line 11: name seems at odds with playbook

ansible/provision-private-proxied-cluster-airgap-deb.yml
```diff
@@ -0,0 +1,65 @@
+### This is not a truly airgapped cluster!!! The intent here is specifically to test the install flow for an airgapped install
```
hcoyote, line 1: Can we change this to something like:

> ### This is a semi-airgapped cluster deployment. The cluster can only access the internet via a proxied connection. This is currently only used for enabling download of Redpanda packages. If you have a centralized proxy, you may substitute that for the configured squid instance.

probably need to identify how to tell how to disable the squid install also

like need to do this for both deb/rpm files

ansible/provision-private-proxied-cluster-airgap-deb.yml
```diff
@@ -0,0 +1,65 @@
...
+
+- name: Bundle DEBs
+  hosts: client
+  vars:
+    basearch: "amd64" # must match the output of `uname -m`. however on ubuntu you need to use the debian version of the output. So for example x86_64 is amd64
+    os_distribution_major_version: "focal" # must match the output of `. /etc/os-release && echo ${VERSION_ID%%.*}`
+    os_distribution: "ubuntu" # must match the output of `. /etc/os-release && echo $ID`
+    redpanda_version: "23.1.13-1" # note that the -1 is the build number for this version. it is typically -1 but if the version is incremented it will be higher
+    rpm_or_deb: "deb"
+    ansible_ssh_common_args: "-o ProxyCommand='ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /app/artifacts/testkey -W %h:%p -q {{ hostvars[groups['client'][0]].ansible_user }}@{{ hostvars[groups['client'][0]].inventory_hostname }}'"
```
hcoyote, line 22: is this only needed for testing? if so, can we figure out how to handle the StrictHostKeyChecking and UserKnownHostsFile just for that? This is going to raise some eyebrows in some secure environments.

ansible/provision-private-proxied-cluster-airgap-deb.yml
```diff
@@ -0,0 +1,65 @@
...
+    enable_tls: true
+    advertise_public_ips: true
+    create_demo_certs: true
+    handle_cert_install: true
+    redpanda_version: latest
+    create_pkg_mgr_proxy: true
+    ca_cert_file: "tls/ca/ca.crt"
+    node_cert_file: "tls/certs/{{ansible_hostname}}/node.crt"
+    cloud_storage_credentials_source: "aws_instance_metadata"
+    ansible_ssh_common_args: "-o ProxyCommand='ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /app/artifacts/testkey -W %h:%p -q {{ hostvars[groups['client'][0]].ansible_user }}@{{ hostvars[groups['client'][0]].inventory_hostname }}'"
```
hcoyote, line 46: same comment as the one above for vars in this file. not actually sure why we define this twice.

ansible/provision-private-proxied-cluster-airgap-rpm.yml
```diff
@@ -0,0 +1,65 @@
...
+    - name: Install and configure proxy on client
+      ansible.builtin.include_role:
+        name: mrlesmithjr.squid
+
+- name: Bundle RPMs
+  hosts: client
+  vars:
+    basearch: "x86_64" # must match the output of `uname -m`
+    os_distribution_major_version: "34" # must match the output of `. /etc/os-release && echo ${VERSION_ID%%.*}`
+    os_distribution: "fedora" # must match the output of `. /etc/os-release && echo $ID`
```
hcoyote, line 19: nit picks: 

use "source" instead of the dot.

also, if we're going to put VERSION_ID in braces, we should put ID in braces as well. I'm actually not sure why we're doing the `%%.*` thing here, this regex is matching every character?

ansible/provision-private-proxied-cluster-airgap-rpm.yml
```diff
@@ -0,0 +1,65 @@
...
+    enable_tls: true
+    advertise_public_ips: true
+    create_demo_certs: true
+    handle_cert_install: true
+    redpanda_version: latest
+    ca_cert_file: "tls/ca/ca.crt"
+    create_pkg_mgr_proxy: true
+    node_cert_file: "tls/certs/{{ansible_hostname}}/node.crt"
+    cloud_storage_credentials_source: "aws_instance_metadata"
+    ansible_ssh_common_args: "-o ProxyCommand='ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /app/artifacts/testkey -W %h:%p -q {{ hostvars[groups['client'][0]].ansible_user }}@{{ hostvars[groups['client'][0]].inventory_hostname }}'"
```
hcoyote, line 46: see ssh comments above

ansible/provision-private-proxied-cluster.yml
```diff
@@ -0,0 +1,51 @@
...
+# root_ca_dir: location of our demo cert authority
+# handle_certs: when false, disables all activity related to our demo CA
+# enable_tls: when true, adds the tls template to the /etc/redpanda/redpanda.yml file
+# advertise_public_ips : causes the cluster to set a public ip on advertised_kafka_api, allowing access to the cluster from the public internet
+---
+- name: Provision client
+  hosts: client
+  vars:
+    squid_acl_localnet:
+      - "10.0.1.0/24"
```
hcoyote, line 17: we should hint to the user on the squid_acl_localnet that they can change to whatever cidr range required and that the redpanda broker nodes would need to be within this range to use the proxy.

hcoyote: (and wherever else we do this with squid_acl_localnet)

hcoyote: I see we hardcode the tf to use this range, so a note pointing them to the TF for that as well.

hcoyote (changes requested):

ansible/provision-private-proxied-cluster.yml
gene-redpanda: @hcoyote  the TF is not hardcoded to use this range. Hard coding in this instance would mean that we set the subnet range inside the module without allowing the user to specify it. In this instance we are setting the range to an arbitrary but standard private IP value in an example, then using it as the starting point for the playbook. The end user could specify different values if they so chose.

ansible/provision-private-proxied-cluster-airgap-deb.yml
gene-redpanda: I would recommend against using it for a secure environment yes, but for ansible it does need to be set -- especially for CI. I really don't think users in secure environments should consume these playbooks without heavy customization. They're intended to demonstrate features and provide a testing bed for CI rather than providing a 1:1 for customers.

ansible/provision-private-proxied-cluster-airgap-deb.yml
gene-redpanda: By way of example, let's say I used this technique to populate the ansible_ssh_args var prior to the tasks requiring it:

```
---
- name: Gather facts for client to enable ssh_args
  hosts: client
  gather_facts: true
  vars:
    ansible_ssh_common_args: "-o ProxyCommand='ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /app/artifacts/testkey -W %h:%p -q {{ hostvars[groups['client'][0]].ansible_user }}@{{ hostvars[groups['client'][0]].inventory_hostname }}'"

# if you want to disable squid install, comment out this task
- name: Provision proxy on client
  hosts: client
  vars:
    squid_acl_localnet:
      - "10.0.1.0/24"
  tasks:
    - name: Install and configure proxy on client
      ansible.builtin.include_role:
        name: mrlesmithjr.squid

# if doing an airgap install, you will only need this if you haven't used a repo management tool, reposync or apt-mirror to sync the Redpanda releases
# if you have done any of the above, you can safely skip this task
- name: Bundle DEBs
  hosts: client
  vars:
    basearch: "amd64" # must match the output of `uname -m`. however on ubuntu you need to use the debian version of the output. So for example x86_64 is amd64
    os_distribution_major_version: "focal" # must match the output of `. /etc/os-release && echo ${VERSION_ID%%.*}`
    os_distribution: "ubuntu" # must match the output of `. /etc/os-release && echo $ID`
    redpanda_version: "23.1.13-1" # note that the -1 is the build number for this version. it is typically -1 but if the version is incremented it will be higher
    rpm_or_deb: "deb"
  tasks:
    - name: Install and configure proxy on client
      ansible.builtin.include_role:
        name: redpanda.cluster.binary_bundler
      run_once: true

## clients[0] is obviously pretty ugly, need to come up with an az friendly way to handle this
## BUT because it is in the playbook it is easy for the user to adjust to their liking at least
- name: Provision proxied private cluster
  hosts: redpanda
  vars:
    https_proxy_value: "{{ hostvars[groups['client'][0]].private_ip }}:3128" # this should be set to your actual proxy value
    rpm_proxy: "{{ hostvars[groups['client'][0]].private_ip }}:3128" # this should be set to your actual proxy value
    root_ca_dir: "{{ playbook_dir }}/tls/ca"
    enable_tls: true
    advertise_public_ips: true
    create_demo_certs: true # in any production scenario this would be disabled
    handle_cert_install: true # this can be commented out if your team has already installed the necessary certs on your broker instances
    redpanda_version: latest
    create_pkg_mgr_proxy: true # this should be commented out if your infrastructure team has already handled
    ca_cert_file: "tls/ca/ca.crt"
    node_cert_file: "tls/certs/{{ansible_hostname}}/node.crt"
    cloud_storage_credentials_source: "aws_instance_metadata"
    enable_airgap: true # if you are using an internal proxy, comment this value out
  tasks:
    - name: Install system prereqs # this can be commented out if your internal team has already handled instance configuration
      ansible.builtin.include_role:
        name: redpanda.cluster.system_setup
    - name: Install and configure CA certs for running tls # this can be commented out if your internal team has already put the necessary certs on the brokers
      ansible.builtin.include_role:
        name: redpanda.cluster.demo_certs
    - name: Install and start redpanda
      ansible.builtin.include_role:
        name: redpanda.cluster.redpanda_broker
      when: not skip_node | default(false) | bool
```

While the fact will be properly specified in the top play, by the time we reach the "Provision proxied private cluster" play it will no longer be specified and the task will hang.

ansible/provision-private-proxied-cluster-airgap-rpm.yml
gene-redpanda:
```
ubuntu@ip-10-0-2-20:~$ source /etc/os-release && echo ${VERSION_ID}
20.04
ubuntu@ip-10-0-2-20:~$ source /etc/os-release && echo ${VERSION_ID%%.*}
20
```

gene-redpanda: IMO . and source are interchangeable, especially for such a small element that is intended to be manually run 🤷‍♀️

gene-redpanda: I'm not using brackets for the second variable because I'm not doing any parameter expansion shenanigans so they aren't needed.

r-vasquez (approved):

PR #201, opened 2023-09-01
aws changes for docs compatibility

gene-redpanda:
https://github.com/redpanda-data/terraform-aws-redpanda-cluster/issues/11

Rather than cleaning up the references I decided to make the default implementation support the desired architecture.


hcoyote:
not sure I understand what you're goal is here. cleaning up references to what?

PR #202, opened 2023-09-01
fix for mdadm issues with larger instance types

gene-redpanda:
https://github.com/redpanda-data/redpanda-ansible-collection/issues/22

Fixes an issue our mdadm implementation has with is4gen.4xlarge instance types

### DIFF
--- a/aws/main.tf
+++ b/aws/main.tf
@@ -2,7 +2,7 @@
 
 module "redpanda-cluster" {
   source                   = "redpanda-data/redpanda-cluster/aws"
-  version                  = "~> 1.0.0"
+  version                  = "~> 1.0.2"
   public_key_path          = var.public_key_path
   broker_count             = var.broker_count
   deployment_prefix        = var.deployment_prefix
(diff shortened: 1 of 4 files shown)


aws/main.tf
```diff
@@ -17,6 +17,10 @@ module "redpanda-cluster" {
   associate_public_ip_addr = var.associate_public_ip_addr
   availability_zone        = var.availability_zone
   client_count             = 1
+  broker_instance_type     = var.instance_type
+  client_instance_type     = var.instance_type
+  prometheus_instance_type = var.instance_type
```
r-vasquez, line 22: Was this removed at some point?

I remember using `instance_type` previously, and even the docs mention it: https://docs.redpanda.com/22.1/deployment/production-deployment-automation/#step-3-create-the-cluster

gene-redpanda: I removed it previously because this is just an example element to show one possible implementation of the module -- and for CI -- and the module actually has three separate instance types defined. But it saves me a little bit of typing and keeps us 1:1 with the docs soo... 😅

r-vasquez (approved):

PR #204, opened 2023-09-11
test: add basic cluster test and add registry bits

r-vasquez:
Registry is enabled by default now, so we want to
test this for now using curl.

Once `rpk registry` is released we need to change
this test to use rpk.

### DIFF
--- a/.buildkite/scripts/test-basic-cluster.sh
+++ b/.buildkite/scripts/test-basic-cluster.sh
@@ -36,6 +36,13 @@ sed 's/$/:9092/' | \
 tr '\n' ',' | \
 sed 's/,$/\n/')
 
+export REDPANDA_REGISTRY=$(sed -n '/^\[redpanda\]/,/^$/p' "${PATH_TO_HOSTS_FILE}" | \
+grep 'private_ip=' | \
+cut -d' ' -f1 |  \
+sed 's/$/:8081/' | \
+tr '\n' ',' | \
+sed 's/,$/\n/')
+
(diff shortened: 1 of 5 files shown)


.buildkite/scripts/test-basic-cluster.sh
```diff
@@ -48,3 +55,6 @@ echo squirrel | "${PATH_TO_RPK_FILE}" topic produce testtopic --brokers "$REDPAN
 
 echo "consuming from topic"
 "${PATH_TO_RPK_FILE}" topic consume testtopic --brokers "$REDPANDA_BROKERS" -v -o :end | grep squirrel || exit 1
+
+echo "testing schema registry"
+for ip_port in $(echo $REDPANDA_REGISTRY | tr ',' ' '); do curl $ip_port/subjects ; done
```
r-vasquez, line 60: The idea here in the future is to use `rpk registry`, once https://github.com/redpanda-data/redpanda/pull/12669 is released. I'll create a follow-up issue after this PR is merged.

gene-redpanda:
@r-vasquez Good start but can we integrate into the existing testing for stable/unstable/instance-size/os?

r-vasquez:
@gene-redpanda I just added commit https://github.com/redpanda-data/deployment-automation/pull/204/commits/2c2419c019e6a152e108cf02dd356ab43bbc6250 which includes testing for TLS in both tiered and TLS tests. :+1:

gene-redpanda (approved):

PR #208, opened 2023-11-01
add basic cluster ci for gcp

gene-redpanda:
Customers seem to prefer the basic playbook to all others when doing initial evaluation so it'd make sense to test the heck out of it.


r-vasquez (approved):
Looks good to me.

PR #232, opened 2024-11-18
bump to FCB 41

gene-redpanda:


aws/main.tf
```diff
@@ -2,8 +2,7 @@
 
 
 module "redpanda-cluster" {
-  source                   = "redpanda-data/redpanda-cluster/aws"
```
andrewstucki, line 5: I would assume you'd want to change this back to an actual published module version prior to merging.

andrewstucki (approved):
This generally looks fine to me.

aws/main.tf
gene-redpanda: yep!

PR #233, opened 2024-11-21
Adding License Testing for Tiered Storage

gene-redpanda:
Adds an operation to apply license separately + testing related to it.

### DIFF
--- a/ansible/operation-apply-license.yml
+++ b/ansible/operation-apply-license.yml
@@ -17,11 +17,13 @@
     - name: Set Redpanda license (string)
       ansible.builtin.command: rpk cluster license set {{ redpanda_license }}
       run_once: true
+      changed_when: false
       when:
         - redpanda_license is defined
 
     - name: Set Redpanda license (path)
       ansible.builtin.command: rpk cluster license set --path {{ redpanda_license_path }}
+      changed_when: false
       run_once: true
       when:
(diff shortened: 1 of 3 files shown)


RafalKorepta (approved):

rpdevmp (approved):
LGTM, 
Just a suggestion to do a squash to combine these

ansible/operation-apply-license.yml
```diff
@@ -0,0 +1,36 @@
...
+- name: Apply Redpanda License
+  hosts: redpanda
+  become: true
+  vars:
+    rpk_bin: rpk
+
+  tasks:
+    - name: Check cluster health
+      ansible.builtin.shell: |
+        {{ rpk_bin }} cluster health | grep -i 'healthy:' | tr -d '[:space:]' | awk -F ':' '{print tolower($2)}'
```
hcoyote, line 11: Probably better to do something like this and check the return code of grep, rather than trying to parse out the true/false status.

rpk cluster health | grep -q -i 'healthy.*false'

returns 0 if one or more lines matched the regex. returns 1 if zero lines matched.

hcoyote: (though, it would be better if rpk cluster health set a return code of 0 or 1 depending on cluster health state, then we could check the return code directly -- don't think that support exists right now)

PR #238, opened 2025-09-29
Move from Make to Task

gene-redpanda:
Updates various clients and tooling in the repo and moves us to task from make to bring the repo into alignment with the k8s operator.


.tasks/connect.yml
```diff
@@ -0,0 +1,150 @@
+version: '3'
+
+vars:
+  ARTIFACT_DIR: "{{.PWD}}/artifacts"
+  DEPLOYMENT_ID: '{{.DEPLOYMENT_ID | default "devex-cicd"}}'
```
chrisseto, line 5: I see this defined and some other variables with the same definitions multiple times (and sometimes not used). Is that intentional? If not would it be better to move share variables into Taskfile.yml at the top level?

.tasks/tools.yml
```diff
@@ -0,0 +1,129 @@
...
+    status:
+      - test -d {{.ANSIBLE_ROLES_PATH}}
+
+  rpk:
+    desc: "Install rpk CLI tool"
+    cmds:
+      - mkdir -p {{.ARTIFACT_DIR}}/tmp
+      - mkdir -p {{.ARTIFACT_DIR}}/bin
+      - |
+        OS=$(uname)
```
chrisseto, line 57: Taskfile has a way to perform this check itself. I don't think I've used it but it may be a bit nicer?

.tasks/tools.yml
```diff
@@ -0,0 +1,129 @@
...
+        if [ "$OS" = "Darwin" ]; then
+          curl -L {{.MAC_RPK}} -o {{.ARTIFACT_DIR}}/tmp/rpk.zip
```
chrisseto, line 59: Just my two cents, I find it better to put non-local variables into the scope of the task so it's clear what is local to this task and what's expected to be shared.

chrisseto (approved):

twmb:
Closing as dormant (214d idle). Reopen or rebase as a new PR if still needed.

PR #239, opened 2025-10-09
Ansible playbooks to manage Redpanda Connect

bfbarkhouse-redpanda:
We don't have much tooling to operate a fleet of self-managed Redpanda Connect nodes, so I created three simple Ansible playbooks to 1.) stage RPCN as a systemd service, 2.) update a pipeline config and restart the service and 3. )stop the service. I've included versions of each playbook for CE and Enterprise. The main difference with Enterprise is it allows you stage a redpanda.license file.


ansible/stage-rpcn-ent.yml
```diff
@@ -0,0 +1,117 @@
...
+      when: ansible_facts['os_family'] == "RedHat"
+
+    - name: Resolve architecture to Redpanda asset suffix
+      set_fact:
+        rpk_arch: "{{ arch_map.get(ansible_architecture, 'amd64') }}"
+
+    - name: Build download URL for rpk zip
+      set_fact:
+        rpk_zip_url: "https://github.com/redpanda-data/redpanda/releases/latest/download/rpk-linux-{{ rpk_arch }}.zip"
+
```
hcoyote: We've got deb/rpm packages for rpk, can we use those instead? This significantly simplifies deployment work and aligns with how we generally deploy on bare-os.

ansible/stage-rpcn-ent.yml
```diff
@@ -0,0 +1,117 @@
...
+        mode: "0755"
+
+    - name: Ensure target directory exists
+      file:
+        path: "{{ rpcn_dir }}"
+        state: directory
+        owner: root
+        group: root
+        mode: "0755"
+
```
hcoyote: This whole section would go away with the os packages for rpk I think

ansible/rpcn_files/redpanda.license
```diff
@@ -0,0 +1 @@
+#Paste license key here
```
hcoyote, line 1: We probably should follow the same license install pattern that's used in redpanda-ansible-collection

ansible/rpcn_files/rpcn-ent.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Redpanda Connect Pipeline
+After=network.target
+
+[Service]
+ExecStart=/opt/rpcn/rpk connect run /opt/rpcn/pipeline.yaml
```
hcoyote: Should this be templated so you can run more than one rpcn pipeline per node?

ansible/rpcn_files/rpcn.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Redpanda Connect Pipeline
+After=network.target
+
+[Service]
+ExecStart=/opt/rpcn/rpk connect run /opt/rpcn/pipeline.yaml
```
hcoyote: same comment here about templating to run more than one pipeline per node

hcoyote (commented):
overall, like the idea. This probably needs to go into redpanda-ansible-collection instead of here. Let's discuss with @gene-redpanda in slack.

PR #241, opened 2025-11-13
Support kc restart

gene-redpanda:
Primary purpose of this is to enable the use inside deployment automation of the new Kafka Connect restart functionality. It also includes various fixes and improvements to CI including moving us off Fedora 39 as our testing platform and aligning the dev tooling to the k8s operator's task from make. 

It also includes a temp commit to allow testing against the correct branch of the ansible collection.


.tasks/certificates.yml
```diff
@@ -0,0 +1,51 @@
...
+    desc: Generate client TLS certificates
+    cmds:
+      - task: client:key
+      - task: client:csr
+      - task: client:crt
+
+  client:key:
+    desc: Generate client private key
+    cmds:
+      - mkdir -p {{.CLIENT_DIR}}
```
RafalKorepta: NIT: This mkdir could be moved to dependency with proper checks. It can be used for all `client:*` depdency.

.tasks/certificates.yml
```diff
@@ -0,0 +1,51 @@
...
+    generates:
+      - "{{.CLIENT_DIR}}/{{.CLIENT_NAME}}.csr"
+    status:
+      - test -f {{.CLIENT_DIR}}/{{.CLIENT_NAME}}.csr
+
+  client:crt:
+    desc: Sign client certificate
+    deps: [client:csr]
+    cmds:
+      - openssl x509 -req -in {{.CLIENT_DIR}}/{{.CLIENT_NAME}}.csr -CA {{.CA_DIR}}/ca.crt -CAkey {{.CA_DIR}}/ca.key -CAcreateserial -out {{.CLIENT_DIR}}/{{.CLIENT_NAME}}.crt -days 365 -sha256
```
RafalKorepta, line 36: How `CA_DIR` is created?

.tasks/certificates.yml
```diff
@@ -0,0 +1,51 @@
...
+    generates:
+      - "{{.CLIENT_DIR}}/{{.CLIENT_NAME}}.crt"
+    status:
+      - test -f {{.CLIENT_DIR}}/{{.CLIENT_NAME}}.crt
+
+  clean:
+    desc: Remove all generated certificates
+    cmds:
+      - rm -rf {{.CA_DIR}}
+      - rm -rf {{.CERT_DIR}}
```
RafalKorepta, line 46: I'm not sure where this variable is used and how this directory is created.

.tasks/cleanup.yml
```diff
@@ -0,0 +1,104 @@
...
+    cmds:
+      - |
+        go run main.go \
+          --project-id "{{.GCP_PROJECT_ID}}" \
+          --prefix "{{.PREFIX}}" \
+          {{if .MIN_AGE}}--min-age "{{.MIN_AGE}}"{{end}} \
+          --auto-approve \
+          --use-gcp-creds-base64
+
+  "all:dry":
```
RafalKorepta, line 86: Why this is quoted?

.tasks/cleanup.yml
```diff
@@ -0,0 +1,104 @@
...
+    desc: "Dry run cleanup for both AWS and GCP"
+    cmds:
+      - task: aws:dry
+      - task: gcp:dry
+
+  "all:ci":
```
RafalKorepta, line 92: Why this is quoted?

.tasks/cleanup.yml
```diff
@@ -0,0 +1,104 @@
...
+    desc: "Cleanup both AWS and GCP in CI mode (auto-approve)"
+    cmds:
+      - task: aws:ci
+      - task: gcp:ci
+
+  default:
+    desc: "Show available cleanup commands"
```
RafalKorepta, line 99: I think you could list task targets using `--help` flag.

ansible/deploy-monitor-tls.yml
```diff
@@ -68,7 +60,7 @@
   roles:
     - grafana.grafana.grafana
   vars:
-    grafana_version: 10.4.1
+    grafana_version: latest
```
RafalKorepta, line 63: I'm not sure why this is not concrete grafana version, but moving target 'latest'?

requirements.yml
```diff
@@ -1,7 +1,9 @@
 collections:
   - name: community.general
-  - name: redpanda.cluster
-    type: galaxy
+  - name: community.grafana
+  - name: https://github.com/redpanda-data/redpanda-ansible-collection.git
+    type: git
+    version: connect-restarts
```
RafalKorepta: As this is temporary how it will be reverted?

RafalKorepta (approved):

requirements.yml
gene-redpanda: I usually just undo and remove the commit so it doesn't pollute the history. It's mostly a hack to keep this repo around 😅

ansible/deploy-monitor-tls.yml
gene-redpanda: Since this repo doesn't get the most love/attention I default to latest whenever possible and then pin if something breaks and I can't get a quick fix together. Otherwise (like with Fedora 39) I end up running so far out of date that I have a lot of additional work to do to bring it into line

.tasks/certificates.yml
gene-redpanda: Yeah there's a whole host of things I want to get done on this taskfile like moving stuff to deps, removing dated tasks I don't need anymore and generally cleaning up the heinous scripts. This was a fast migration as a first and foremost 😅

.tasks/certificates.yml
gene-redpanda: The certs and directory structure are all created by this role roles/demo_certs so that end users/CS have an easy button for tiered storage cluster deploys. I need a cleanup option that others don't because if I run enough tests I start having collisions 😂

.tasks/cleanup.yml
gene-redpanda: I have issues with cross calls of tasks so whenever I run into one I quote it and the issue stops

.tasks/certificates.yml
gene-redpanda: My intent is to do a follow on PR to improve a lot of this stuff the next time I work in this repo

david-yu:
Is this good to merge?

PR #248, opened 2026-05-15
Use redpanda-ansible-collection git branch for GCP AR URL testing

PrzemekZglinicki:
Point redpanda.cluster collection at the devprod-4109-gcp-ar-urls branch instead of Galaxy, so the Buildkite pipeline validates GCP AR URLs before the collection is published. Also removes ansible/group_vars/all.yml which was a temporary URL override now superseded by the collection change.

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,9 +1,7 @@
 collections:
   - name: community.general
   - name: redpanda.cluster
-    type: git
-    source: https://github.com/redpanda-data/redpanda-ansible-collection.git
-    version: devprod-4109-gcp-ar-urls
+    type: galaxy
   - name: ansible.posix
   - name: grafana.grafana
     version: 5.6.0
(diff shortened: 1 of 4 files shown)


requirements.yml
```diff
@@ -1,7 +1,9 @@
 collections:
   - name: community.general
   - name: redpanda.cluster
-    type: galaxy
+    type: git
+    source: https://github.com/redpanda-data/redpanda-ansible-collection.git
+    version: devprod-4109-gcp-ar-urls
```
ivotron: shouldn't this be dropped?

ivotron (changes requested):

requirements.yml
PrzemekZglinicki: It is already dropped :D

PrzemekZglinicki:
https://buildkite.com/redpanda/deployment-automation/builds/424/list - related successful bk pipeline

ivotron (approved):

